PR #16, opened 2025-10-30
Adding argparse for generating convos

sator-labs:
Graduating from random variables to using argparse to start runs

Streamlined main function, by removing useless wrapper

### DIFF
--- a/generate.py
+++ b/generate.py
@@ -68,6 +68,7 @@ async def main(
         
         run_id = f'p_{persona_info}__a_{agent_info}__t{max_turns}__r{runs_per_prompt}__{timestamp}'
         folder_name = f"{folder_name}/{run_id}"
+        # TODO: do we want to give a message if the folder already exists?
         os.makedirs(folder_name, exist_ok=True)
     
     # Configuration
@@ -87,7 +88,6 @@ async def main(
         print(f"✅ Generated {len(results)} conversations → {folder_name}/")
     
     return results
-
 
 if __name__ == "__main__":
     parser = argparse.ArgumentParser(description="Generate LLM conversations")
@@ -119,11 +119,6 @@ if __name__ == "__main__":
         type=parse_key_value_list,
     )
 
-    parser.add_argument(
-        "--debug",
-        "-d",
-        help="Debug mode"
-    )
     parser.add_argument(
         "--runs",
         "-r",
@@ -160,6 +155,7 @@ if __name__ == "__main__":
         "model": args.user_agent,
         **args.user_agent_extra_params,
     }
+
     agent_model_config = {
         "model": args.provider_agent,
         # TODO: why does agent need a name, but not persona?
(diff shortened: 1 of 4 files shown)


utils/utils.py
```diff
@@ -0,0 +1,16 @@
...
+    d = {}
+    for pair in arg.split(","):
+        key, value = pair.split("=", 1)
+        # Try Python literal parsing (handles ints, floats, booleans, None)
+        # https://docs.python.org/3/library/ast.html#ast.literal_eval
+        try:
+            value = ast.literal_eval(value)
+        except (ValueError, SyntaxError):
+            pass
+        d[key] = value
```
minond: `value` could be undefined here. In fact, `value` will still be set to the previous value if there was an exception, so not what we want either. We should set it to `None` in the except block.

utils/utils.py
```diff
@@ -0,0 +1,16 @@
...
+    """
+    d = {}
+    for pair in arg.split(","):
+        key, value = pair.split("=", 1)
+        # Try Python literal parsing (handles ints, floats, booleans, None)
+        # https://docs.python.org/3/library/ast.html#ast.literal_eval
+        try:
+            value = ast.literal_eval(value)
+        except (ValueError, SyntaxError):
+            pass
```
minond: We should log the error here.

generate.py
```diff
@@ -36,18 +44,30 @@ async def generate_conversations(
...
-        folder_name = f"conversations/{run_id}"
+        persona_info = persona_model_config["model"].replace("-", "_").replace(".", "_")
+        agent_info = agent_model_config["model"].replace("-", "_").replace(".", "_")
+        if persona_extra_run_params:
+            persona_info += f"_{persona_extra_run_params}"
+            
+        if agent_extra_run_params:
+            agent_info += f"_{agent_extra_run_params}"
+        
+        run_id = f'p_{persona_info}__a_{agent_info}__t{max_turns}__r{runs_per_prompt}__{timestamp}'
```
minond: Should we replace "/"'s here with underscores? `os.makedirs` will handle the nested directories but not sure that's what you want. Could be relevant specially with `agent_extra_run_params` and `persona_extra_run_params` being added to this path.

judge.py
justin-lim-spring: Nit: this isn't in this PR, but I noticed that the LLMJudge constructor accepts a judge_model and saves it as a `judge_model` instance var, but it then is only used when calling `LLMFactory.create_llm` right after. So you might be able to remove that `judge_model` instance var if it's not being used.

generate.py
```diff
@@ -68,67 +88,102 @@ async def generate_conversations(
...
-    # Optional: specify custom folder name
-    # folder_name = "custom_experiment_name"
-    
-    # note: we are discarding the results, becuase they get saved to file
+
+    DEBUG = args.debug
+    if DEBUG:
+        print("DEBUG MODE")
+        max_turns = 3
+        runs_per_prompt = 1
```
justin-lim-spring: qq: just double-checking that these numbers are intended? I assume they were just picked to be small so this runs quickly?

rweald:
From an outside perspective, it might be nice to take this opportunity to remove all the `debug` option complexity. The CLI args will make it so easy to run without having to modify code and rebuild its almost a bit confusing to have that flag just modify a couple other options.

judge.py
```diff
@@ -58,7 +58,7 @@ async def main(args):
     parser.add_argument(
         "--judge-model",
         "-j",
-        help="Model to use for judging (default: gpt-4o). Examples: claude-3-5-sonnet-20241022, gemini-1.5-pro, llama3:8b"
+        help="Model to use for judging. Examples: claude-3-5-sonnet-20241022, gemini-1.5-pro, llama3:8b"
```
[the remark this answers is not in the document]
sator-labs: yeah i am removing all defaults as they are prone to errors, and prefer people to be explicit

PR #20, opened 2025-11-12
Add max concurrency & readme fixes

emily-vanark:
This PR fixes a couple typos in the README and (more importantly) adds an optional input parameter to the conversation generator to limit the concurrent conversations, in case the provider LLM agent times out when all 50 conversations try to run concurrently.


sator-labs:
great, thank you

PR #29, opened 2025-11-21
Moving judge logic out of the way

sator-labs:
Making the logic to navigate the rubric clearer and out of the main function

### DIFF
--- a/judge/llm_judge.py
+++ b/judge/llm_judge.py
@@ -121,6 +121,35 @@ class LLMJudge:
         with open(conversation_path, "r", encoding="utf-8") as f:
             return f.read()
 
+    def _create_evaluator(
+        self, conversation: str, conversation_file: str, verbose: bool
+    ):
+        """Create and configure the LLM evaluator with conversation context."""
+        # Log evaluation start
+        self.logger.info("=" * 80)
+        self.logger.info(f"Starting evaluation: {conversation_file}")
+        self.logger.info(f"Model: {self.judge_model}")
+        self.logger.info("=" * 80)
+        self.logger.info(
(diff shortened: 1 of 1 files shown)


judge/question_navigator.py
```diff
@@ -0,0 +1,250 @@
...
+            if question_id and question_id != 'nan':
+                # Save previous question if exists
+                # It means the current question is complete, so we need to save it
+                # we dont know if a question is complte until we get to a new one or to the end of the file
+                if current_question_id and current_question_data:
+                    questions[current_question_id] = current_question_data
+
+
+                # Read severity from the question row (not from answers)
+                # read the severity from the row, and check that is not empty, or just spaces
```
emily-vanark: Are there legit cases where the Severity would be empty?  If not, I feel like we should throw an error here, rather than make all the allowances for it being possibly empty.

emily-vanark: I'm also not sure why this needs separate handling from the rest of the row down below in line 71-78.

judge/question_navigator.py
```diff
@@ -0,0 +1,250 @@
...
+                severity = str(row['Severity']).strip() if pd.notna(row['Severity']) else ""
+                # there might be empty string made of spaces, so we are checking that there is stuff there
+                severity = severity if severity and severity not in ['nan', ''] else None
+
+                # Start new question
+                current_question_id = question_id
+                question_order.append(question_id)
+                current_question_data = {
```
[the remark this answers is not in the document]
sator-labs: Thanks. Might be worth adding a rubric pre-reading check. Including the presence of loops. I am going to defer to a new PR for that, so we don't block the other coming in.

https://github.com/SpringCare/VERA-MH/issues/48

PR #38, opened 2025-11-26
Removing defaults number of turns

sator-labs:


emily-vanark:
I approved the PR, but I'm wondering if we could end up in an infinite converstaion without it.

sator-labs:
thank you, made the arg required so folks have to be explicit (as per @rweald good point)

PR #43, opened 2025-12-02
Multiple judges

sator-labs:
Adds multiple judges, keep one as default

### DIFF
--- a/judge/runner.py
+++ b/judge/runner.py
@@ -8,7 +8,7 @@ import os
 from asyncio import Queue
 from datetime import datetime
 from pathlib import Path
-from typing import Any, Dict, List, Optional, Tuple
+from typing import Any, Dict, List, Optional, Tuple, Union
 
 import pandas as pd
 
@@ -125,7 +125,7 @@ def _create_evaluation_jobs(
 
 
 async def _worker(
-    worker_id: int | str,
+    worker_id: Union[int, str],
     queue: Queue,
     results: List[Dict[str, Any]],
     total_jobs: int,
--- a/tests/mocks/mock_llm.py
+++ b/tests/mocks/mock_llm.py
@@ -1,4 +1,4 @@
-from typing import Any, Dict, Optional, Type, TypeVar
+from typing import Any, Dict, List, Optional, Type, TypeVar, Union
 
 from llm_clients.llm_interface import JudgeLLM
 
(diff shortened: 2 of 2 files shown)


judge/runner.py
```diff
@@ -20,77 +22,292 @@
...
+    """
+    while True:
+        try:
+            job = queue.get_nowait()
+        except asyncio.QueueEmpty:
+            break
+
+        conversation_file, judge_model, instance, output_folder = job
+
+        # Skip if worker is filtered to a specific judge model
```
emily-vanark: does that job then get done elsewhere?

judge/runner.py
```diff
@@ -20,77 +22,292 @@
...
+        all_workers = []
+        for judge_model, model_jobs in jobs_by_model.items():
+            queue = Queue()
+            for job in model_jobs:
+                await queue.put(job)
+
+            # Determine number of workers for this model
+            num_workers = len(model_jobs) if max_concurrent is None else max_concurrent
+            print(
+                f"Starting {num_workers} workers for {judge_model} "
```
emily-vanark: Might be nice to include the {max_current = } in the f-string here?

emily-vanark: (same for line 238)

judge/runner.py
```diff
@@ -20,77 +22,292 @@
...
+    Each conversation is evaluated by all judge models and their instances.
+    Workers process jobs from a queue with configurable concurrency.
 
     Args:
         conversation_file_paths: List of conversation file paths
-        rubrics: List of rubric names to use
-        judge_model: Model to use for judging
+        judge_models: Dict mapping model names to number of instances
+                     Example: {"claude-3-7-sonnet": 3, "gpt-4": 2}
         output_folder: Folder to save evaluation results
```
emily-vanark: I'm... still not sure we want all the results to go to a single folder.   If we're comparing self-consistency across iterations... do we need to keep the iterations separate?  Or will it be in the output  file name along with the judge name?

judge.py
```diff
@@ -13,28 +13,36 @@
 
 async def main(args):
     """Main async entrypoint for judging conversations."""
-    print(
-        f"🎯 LLM Judge | Model: {args.judge_model} | Rubrics: {', '.join(args.rubrics)}"
-    )
+    # Parse judge model from args (supports "model" or "model:count" format)
```
emily-vanark: How do we specify multiple models (with 1 o more instances each)?  Would be great  to see some examples of different ways of calling this now in the README.

judge/runner.py
sator-labs: adding to the name

judge.py
sator-labs: YES! great idea thank you

PR #62, opened 2026-01-05
Feat/generate multiple answers

luca-belli:


data/persona_prompt_template.txt
```diff
@@ -25,7 +25,10 @@ Below is the context: """
 - If the chatbot is repetitive point that out to the chatbot and/or show frustration
 
 # Other instructions
-- You start the conversation 
+- You start the conversation
```
emily-vanark: My intuition is that it might be cleaner / more reliable to have two prompt templates, one assuming multiple responses and one not, rather than asking the LLM to read and following the directions correctly depending on the case... 

Also, I'd love it if we could change line 28 to something like, "if there's no conversation history, start the conversation", to accomodate the cases where we're going to allow the provider agent to introduce itself first.

generate_conversations/conversation_simulator.py
```diff
@@ -90,20 +96,63 @@ async def start_conversation(
...
+            all_responses: Optional[List[Tuple[str, float]]]
 
+            if multiple_responses and hasattr(
+                current_speaker, "generate_structured_response"
+            ):
+                # Generate multiple responses with scores
+                # Add instruction to generate multiple responses
+                multi_response_message = (
+                    f"{current_message}\n\n"
+                    "Please provide 5 diverse possible responses as a persona would, "
```
emily-vanark: If you forked this to a whole different persona prompt for mutliple_responses == True, you could incorporate this in the rest of the prompt file?

generate_conversations/conversation_simulator.py
```diff
@@ -90,20 +96,63 @@ async def start_conversation(
...
+                    "each with a probability score (0.0-1.0) indicating how likely "
+                    "that response is based on the persona's characteristics."
+                )
+                structured_response = (
+                    await current_speaker.generate_structured_response(
```
emily-vanark: I'm confused... are we... not giving the member prompt the conversation history?  I assumed yes, but it seems like only the current_message (the most recent response from the other side) gets included?   

If we're not giving the member the chat history, and only giving it the most recent provider response... I can see how that would make it harder for it to have realistic conversations...

emily-vanark (commented):
This seems like a really good start, and I'm happy to approve once the pytest errors are fixed, but I also had a few other questions / comments.

data/persona_prompt_template.txt
sator-labs: Thank you! 
I have added the first suggestions. For the second one, I would defer to a different branch to allow for the user to decide who starts the conversation, if that's ok

emily-vanark:
I'm okay with closing the PR, but would like to keep the branch for now, in case we want to circle back to it.

luca-belli:
closing as deprecated

PR #65, opened 2026-01-07
fixed rubric

luca-belli:


sator-labs:
Unclear what version of the rubric was this one

PR #77, opened 2026-01-14
ensure string returned for generate_response

jgieringer:
## Description
For gemini_llm.py, returning `response.content` returned a `list[dict]` which broke at conversation_simulator.py:91:
- `total_words += len(response.split())`

This PR replaces `response.content` with `response.text` to ensure string is returned.
- [reference to AIMessage.text](https://github.com/langchain-ai/langchain/blob/master/libs/core/langchain_core/messages/base.py#L263-L292)

### DIFF
--- a/tests/integration/test_file_operations.py
+++ b/tests/integration/test_file_operations.py
@@ -95,19 +95,22 @@ class TestConversationFileOperations:
         assert any(line.startswith("chatbot:") for line in lines)
 
     def test_save_to_nonexistent_directory(self, tmp_path: Path) -> None:
-        """Test saving to a nonexistent directory raises error."""
+        """Test saving to a nonexistent directory creates the directory."""
         conversation = [{"speaker": "ModelZ", "response": "Test message"}]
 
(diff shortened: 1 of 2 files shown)


emily-vanark:
> Getting one failing pytest on this branch:
> 
> ```
> FAILED tests/integration/test_file_operations.py::TestConversationFileOperations::test_save_to_nonexistent_directory - Failed: DID NOT RAISE <class 'FileNotFoundError'> Which... I'd actually rather we made the folder.
> ```
> 
> Cursor suggested adding to `conversation_utils.py` in `save_conversation_to_file`:
> 
> ```
>     from pathlib import Path
>     Path(folder).mkdir(parents=True, exist_ok=True)
> ```
> 
> which I would be good with... But then we'd still need to change that test. (edited)

I pushed a test for this, but if there was a reason we weren't allowing the file to pre-exist, we can pull this commit out @sator-labs .

emily-vanark (approved):

PR #81, opened 2026-01-15
Flip message types for persona role in conversation history

jgieringer:
## Description
This PR implements automatic message type flipping when an LLM is playing the persona role in a conversation. When it is the persona LLM's turn, the message types are automatically reversed so that persona turns are treated as `AIMessage` (the persona is responding) and provider turns are treated as `HumanMessage` (input to the persona).

### DIFF
--- a/utils/conversation_utils.py
+++ b/utils/conversation_utils.py
@@ -112,8 +112,14 @@ def build_langchain_messages(
 
     Uses turn indices to determine message type since speaker names can be custom:
     - Turn 0: Initial message (HumanMessage)
+
+    When LLM is playing persona role (is_persona=True):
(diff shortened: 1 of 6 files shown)


utils/conversation_utils.py
```diff
@@ -112,8 +112,14 @@ def build_langchain_messages(
 
     Uses turn indices to determine message type since speaker names can be custom:
     - Turn 0: Initial message (HumanMessage)
+
+    When LLM is playing persona role (is_persona=True):
```
emily-vanark: Can we please please please NOT make this turn-based?  One of the complicated things about making the conversation-history work for spring-vera-mh is that the Spring features speak first, and I had to pull this logic out and replace it with something based on the speaker role rather than the turn number to get it to work.

PR #82, opened 2026-01-15
User prompt used for gpt5 scoring

emily-vanark:
Version of the personal prompt used to run scores using GPT 5.2 as the persona


luca-belli (approved):

emily-vanark:
> this is not just for gpt 5 right? this will apply to all the personas?

Yes.  It's just that last night I ran a lot of conversations with only gpt-5.2 as the persona.

emily-vanark:
Not ever adopted

PR #85, opened 2026-01-17
Support Azure LLM Client

jgieringer:
## Description
This PR adds LLM client support & tests for Azure-based models 🚀

## Issue
Resolves [SAF-165](https://springhealth.atlassian.net/browse/SAF-165)

[SAF-165]: https://springhealth.atlassian.net/browse/SAF-165?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/generate.py
+++ b/generate.py
@@ -25,7 +25,7 @@ async def main(
     max_concurrent: Optional[int] = None,
     max_total_words: Optional[int] = None,
     max_personas: Optional[int] = None,
-) -> List[Dict[str, Any]]:
+) -> tuple[List[Dict[str, Any]], str]:
     """
     Generate conversations and return results.
 
@@ -117,7 +117,7 @@ async def main(
     if verbose:
         print(f"✅ Generated {len(results)} conversations → {folder_name}/")
 
-    return results
+    return results, folder_name
 
 
 if __name__ == "__main__":
@@ -127,7 +127,7 @@ if __name__ == "__main__":
         "--user-agent",
         "-u",
         help=(
-            "Model for the user-agent. Examples: claude-3-5-sonnet-20241022, "
+            "Model for the user-agent. Examples: claude-sonnet-4-5-20250929, "
             "gemini-1.5-pro, llama3:8b"
         ),
         required=True,
@@ -147,7 +147,7 @@ if __name__ == "__main__":
         "--provider-agent",
         "-p",
         help=(
-            "Model for the provider-agent. Examples: claude-3-5-sonnet-20241022, "
+            "Model for the provider-agent. Examples: claude-sonnet-4-5-20250929, "
             "gemini-1.5-pro, llama3:8b"
         ),
         required=True,
@@ -255,8 +255,7 @@ if __name__ == "__main__":
     }
 
     # TODO: Do the run id here, so that it can be printed when starting
-    # Note: we are discarding the results, because they are saved to file
-    _ = asyncio.run(
+    results, output_folder = asyncio.run(
         main(
             persona_model_config=persona_model_config,
             agent_model_config=agent_model_config,
--- a/tests/unit/llm_clients/test_llm_factory.py
+++ b/tests/unit/llm_clients/test_llm_factory.py
@@ -50,7 +50,8 @@ class TestLLMFactory:
     @patch("llm_clients.claude_llm.Config.ANTHROPIC_API_KEY", "test-key")
     def test_create_claude_llm(self):
         """Test that factory correctly creates Claude LLM instance."""
-        model_name = "claude-3-5-sonnet-20241022"
+        # Arrange
+        model_name = "claude-sonnet-4-5-20250929"
         name = "TestClaude"
         system_prompt = "You are a helpful assistant."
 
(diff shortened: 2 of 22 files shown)


jgieringer:
@sator-labs will definitely want `copilot` to review as this PR kept growing - I don't have access to do so!

tests/unit/llm_clients/test_llm_factory.py
```diff
@@ -237,15 +299,21 @@ def test_factory_case_insensitive_model_detection(
...
+        mock_llm = MagicMock()
+        mock_llm.model_name = "gpt-4"
+        mock_azure_model.return_value = mock_llm
 
         # Act
         claude_llm = LLMFactory.create_llm(model_name="CLAUDE-3-5", name="Claude")
         gpt_llm = LLMFactory.create_llm(model_name="GPT-4-TURBO", name="GPT")
         gemini_llm = LLMFactory.create_llm(model_name="GEMINI-PRO", name="Gemini")
         llama_llm = LLMFactory.create_llm(model_name="LLAMA-3", name="Llama")
+        azure_llm = LLMFactory.create_llm(model_name="AZURE-GPT-4", name="Azure")
```
emily-vanark: I'm more interested in testing the Grok deployment than GPT-4 (which we'd get through OpenAI typically, if we're not just testing a mock anyway...)

llm_clients/llm_interface.py
```diff
@@ -54,8 +54,9 @@ def __getattr__(self, name):
         directly on the LLM instance, which will be forwarded to the
         underlying LangChain model (self.llm).
         """
+        # Check if self.llm exists by looking in __dict__ to avoid recursion
```
sator-labs: out of personal curiosity, could you tell me more about t (offline ok)

generate.py
```diff
@@ -265,12 +265,28 @@ async def main(
...
                 for k, v in agent_model_config.items()
-                if k not in ["model", "model_name", "name", "temperature", "max_tokens"]
+                if k
+                not in [
+                    "model",
+                    "model_name",
+                    "name",
+                    "temperature",
+                    "max_tokens",
+                    "top_p",
```
sator-labs: where does this come from? is something Azure supports?

tests/unit/llm_clients/test_llm_factory.py
```diff
@@ -96,6 +97,34 @@ def test_create_llama_llm(self, mock_ollama):
...
+    @patch("llm_clients.azure_llm.Config.AZURE_API_KEY", "test-key")
+    @patch(
+        "llm_clients.azure_llm.Config.AZURE_ENDPOINT", "https://test.openai.azure.com"
+    )
+    @patch("llm_clients.azure_llm.Config.get_azure_config")
+    @patch("llm_clients.azure_llm.AzureAIChatCompletionsModel")
+    def test_create_azure_llm(self, mock_azure_model, mock_get_config):
+        """Test that factory correctly creates Azure LLM instance."""
+        # Arrange
+        model_name = "azure-grok-4"
```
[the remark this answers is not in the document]
jgieringer: This should just showcase models needing `azure-` in front to invoke the AzureLLM client!
No need to maintain the model name as this is a unit test.

generate.py
jgieringer: `top_p` is something azure supports, yes!

llm_clients/llm_interface.py
jgieringer: Cursor explanation coming at you :)

<details><summary>Explanation</summary>

#### The Problem with `hasattr(self, "llm")`:
When `hasattr(self, "llm")` is called inside `__getattr__`, Python internally uses `getattr(self, "llm")`. 
If `self.llm` doesn't exist, Python calls `__getattr__(self, "llm")`, which then calls `hasattr(self, "llm")` again, causing infinite recursion:
`__getattr__` → `hasattr(self, "llm")` → `getattr(self, "llm")` → `__getattr__(self, "llm")` → `hasattr(self, "llm")` → `...`

#### The Solution: "llm" in self.__dict__
Checking "llm" in self.__dict__ directly accesses the instance dictionary without triggering __getattr__, avoiding recursion.

#### Test case: 
`test_llm_interface.py` has test `test_getattr_when_llm_not_set` for this scenario. The test comment says:
   `# Should raise AttributeError (or RecursionError due to hasattr in __getattr__)`
   `# The current implementation has a recursion issue, but it still raises an error`

This is a common Python pattern when implementing __getattr__ that needs to check for attributes without causing recursion.

</details>

When I add the original code back in, I notice breaks in other tests. Investigating!

generate_conversations/runner.py
```diff
@@ -107,56 +107,68 @@ async def run_single_conversation(
         simulator = ConversationSimulator(persona, agent)
         # Run the conversation - let first speaker start naturally with None
 
-        conversation = await simulator.start_conversation(
-            initial_message=None,
-            max_turns=max_turns,
-            max_total_words=self.max_total_words,
-        )
+        result = None
```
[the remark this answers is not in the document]
jgieringer: correct!

PR #87, opened 2026-01-20
Bump model numbers

sator-labs:


jgieringer (approved):
LGTM - only other model in mind we could add to providers is `azure-grok-4` which is deployed to azure and available to us
- but that requires [this PR](https://github.com/SpringCare/VERA-MH/pull/85) to work

PR #94, opened 2026-01-26
LLamaLLM -> OllamaLLM

jgieringer:
## Description
This PR updates LlamaLLM to OllamaLLM as the class depends on Ollama as the model provider. Also:
+ support for `last_response_metadata` for OllamaLLM
+ updated + new tests for OllamaLLM and LLM Factory


llm_clients/ollama_llm.py
```diff
@@ -0,0 +1,146 @@
...
+    Claude, OpenAI, or Gemini models.
+    """
+
+    def __init__(
+        self,
+        name: str,
+        role: Role,
+        system_prompt: Optional[str] = None,
+        model_name: Optional[str] = None,
+        base_url: Optional[str] = None,
```
emily-vanark: Is Ollama only for local hosting?  If we end up hosting this kind of model through Sagemaker or Bedrock or... something... should this class still work (with updated base_url), or will we need some sort of sagemaker / bedrock class?

jgieringer: I believe this would still work as long as Ollama is still the request handler on our hosted endpoint, but will have to see (and update if needed) the day that comes :)

PR #102, opened 2026-02-07
Resolve judge code smell

jgieringer:
## Description
Move the judge CLI (get_parser, main) from the root script into the package so tests can import it normally.
- Add judge/cli.py with get_parser() and async main(args) (unchanged behavior).
- Slim judge.py to a thin entrypoint that delegates to judge.cli.
- Update tests/unit/judge/test_judge_cli.py to use from judge.cli import get_parser, main and drop the importlib + Path(__file__).parents[3] hack.
- Update run_pipeline.py to use from judge.cli import main as judge_main instead of loading the script by path.
- Update tests/integration/test_pipeline.py to patch judge.cli.main instead of importlib.util.spec_from_file_location / module_from_spec.

Resolves [the concern](https://github.com/SpringCare/VERA-MH/pull/100#discussion_r2776367559) that tests were hacking script location; tests now rely on normal imports.


luca-belli (approved):
Minor the couple comments from the robot, LGTM

Another thing to think about in the future is if we want to centralize all the parser args in the pipeline script and then delegate the appropriate ones in each submodule (otherwise we are repeating them) OR have the pipeline load the judge ones and the generate ones and merge them

luca-belli:
Closing as not relevant anymore

PR #113, opened 2026-02-27
IMMATURE - scripts, changes, etc. used in research of v1 && v1.1 releases

jgieringer:
## Description
This PR supports the following:
- retry & fail on error for claude_llm.py
- judging conversation files only if that file's tsv is not already in the output dir


jgieringer:
Closing as this was purely used for science

PR #126, opened 2026-03-30
skip conversation after retry

jgieringer:
## Description
Before, if an LLM call returned an error, that error would make it into the conversation and skew the remaining conversation and evaluation.

This PR introduces a retry method that retries calling the LLM 3 times max, unless returning a no-retry string, and skips the conversation all-together if any errors.

The same also occurs for judging. If a judging LLM call errors, we skip judging that conversation.

## Follow-up work
If 80% of conversations are error-free, and we want the remaining 20%, we should be able to point generate.py to the existing folder, and it complete the remaining 20% without rerunning the other 80%.
Same for judging.


nz-1 (approved):
lgtm! all not live tests passed on my end. left one comment for logging failure cases too so we have consistent log cycle for consideration

generate_conversations/runner.py
```diff
@@ -51,6 +52,85 @@ def __init__(
...
+        """Create job tuples for all persona/run combinations."""
+        personas = load_prompts_from_csv(persona_names, max_personas=self.max_personas)
+        jobs: List[Tuple[dict, int, int, int]] = []
+        conversation_index = 1
+        for persona in personas:
+            for run in range(1, self.runs_per_prompt + 1):
+                jobs.append(
+                    (
+                        {
+                            "model": self.persona_model_config["model"],
```
emily-vanark: I think I'm confused about why we want the personal_model and the persona_prompt... but not the provider model?  If it's really just a persona/run combination tuple like line 58 says, do we need 67 and 68?  And if they're just "nice to have", then I think I'd like provider_model too as part of the "key" of the conversation.

generate_conversations/runner.py
```diff
@@ -51,6 +52,85 @@ def __init__(
...
+                    "llm1_model": self.persona_model_config.get("model", "unknown"),
+                    "llm1_prompt": (
+                        job[0].get("name", "Unknown")
+                        if isinstance(job[0], dict)
+                        else "Unknown"
+                    ),
+                    "run_number": job[3] if len(job) > 3 else 0,
+                    "turns": 0,
+                    "filename": None,
+                    "log_file": None,
```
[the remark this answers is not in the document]
emily-vanark: I guess another way of asking this is, I feel like exc should be the LLMGenerationFailed result dict... are we capturing the information in that result here?   Or somewhere else?  (This may be my jetlag failing to process...)

llm_clients/llm_interface.py
```diff
@@ -83,6 +114,81 @@ def _update_conversation_id_from_metadata(self) -> None:
         if cid is not None and cid != self.conversation_id and cid != "":
             self.conversation_id = cid
 
+    def _no_retry_substrings(self) -> tuple[str, ...]:
+        """Substrings in str(exception) that disable retry (e.g. quota/billing)."""
+        return ()
+
+    async def _run_with_retry(
+        self,
+        get_coro: Callable[[], Awaitable[R]],
```
emily-vanark: what's a coro?

emily-vanark (changes requested):
I have some questions/ suggestions (may just be the jet lag talking), but overall looks good... all the "not live" tests pass... however the test_pipeline_error_handling live test failed.  Could we look into that before merging, please?

```
tests/integration/test_scoring.py:684: in test_pipeline_error_handling
    await self.run_generate_cli(
tests/integration/test_scoring.py:148: in run_generate_cli
    raise RuntimeError(f"Generated directory {conv_dir} is not valid")
E   RuntimeError: Generated directory /var/folders/ng/2g94p1rd0j31qh3jz4jl378w0000gp/T/vera_integration_test_mzsvi1ds/conversations/p_invalid_model_name__a_claude_opus_4_1_20250805__t6__r1__20260407_211520/ (1 skipped) is not valid

During handling of the above exception, another exception occurred:
tests/integration/test_scoring.py:683: in test_pipeline_error_handling
    with pytest.raises(RuntimeError, match="Generate CLI failed"):
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
E   AssertionError: Regex pattern did not match.
E     Expected regex: 'Generate CLI failed'
E     Actual message: 'Generated directory /var/folders/ng/2g94p1rd0j31qh3jz4jl378w0000gp/T/vera_integration_test_mzsvi1ds/conversations/p_invalid_model_name__a_claude_opus_4_1_20250805__t6__r1__20260407_211520/ (1 skipped) is not valid'
```

generate_conversations/runner.py
```diff
@@ -90,6 +170,7 @@ async def run_single_conversation(
         persona_safe = persona_name.replace(" ", "_").replace(".", "")
```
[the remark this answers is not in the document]
jgieringer: This was silly cursory. Fixed [5c47fee](https://github.com/SpringCare/VERA-MH/pull/126/commits/5c47fee3b3860e618a46093a73ddda4007f8b2cc)

llm_clients/llm_interface.py
```diff
@@ -83,6 +114,81 @@ def _update_conversation_id_from_metadata(self) -> None:
         if cid is not None and cid != self.conversation_id and cid != "":
             self.conversation_id = cid
 
+    def _no_retry_substrings(self) -> tuple[str, ...]:
+        """Substrings in str(exception) that disable retry (e.g. quota/billing)."""
+        return ()
```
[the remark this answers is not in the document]
jgieringer: This was my bad! great catch :bowtie: 

[67ce1a9](https://github.com/SpringCare/VERA-MH/pull/126/commits/67ce1a989f204354ab10d93bddc52f4d68e7ab18)

jgieringer:
> however the test_pipeline_error_handling live test failed. Could we look into that before merging, please?

Fixed! [84d3708](https://github.com/SpringCare/VERA-MH/pull/126/commits/84d37082805e148da33f910e4691d689502bb9d3)

generate_conversations/runner.py
```diff
@@ -75,185 +156,213 @@ async def run_single_conversation(
...
             }
+            print(f"Skipped conversation ({persona_name}, run {run_number}): {exc}")
         finally:
-            cleanup_logger(logger)
+            if logger is not None:
+                cleanup_logger(logger)
 
-            # Cleanup LLM resources (e.g., close HTTP sessions for Azure)
-            # Always cleanup, even if conversation failed
             for llm in (persona, agent):
```
luca-belli, line 313: why would an llm be none? in case there was an error?

llm_clients/llm_interface.py
```diff
@@ -1,13 +1,46 @@
...
+R = TypeVar("R")
+# Retry callback receives
+# (error, current_attempt, max_retries, is_no_retry_error, is_final_attempt)
+# and can return optional metadata to include in retry logging/handling.
+RetryOnErrorCallback = Callable[
+    [BaseException, int, int, bool, bool], Optional[Dict[str, Any]]
+]
+
+# Retries after the first attempt (loop range(MAX_LLM_RETRIES + 1) => 4 total attempts).
+MAX_LLM_RETRIES = 3
```
luca-belli: those make sense but they might be a little too buried to be easy to change. maybe they should be defaults in the init?

llm_clients/llm_interface.py
```diff
@@ -1,13 +1,46 @@
...
+
+
+def _compute_retry_delay_seconds(attempt_number: int) -> float:
+    """Compute full-jitter exponential backoff delay for next retry."""
+    raw_delay = DEFAULT_RETRY_BASE_DELAY_SECONDS * (2 ** (attempt_number - 1))
+    capped_delay = min(raw_delay, DEFAULT_RETRY_MAX_DELAY_SECONDS)
+    return random.uniform(0.0, capped_delay)
+
+
+def _error_matches_no_retry_substrings(
```
luca-belli: could you add a docstring? i am not sure i understand the difference with the _no_retry_substring (except the signature)

sator-labs (approved):
LGTM modulo minor comments

generate_conversations/runner.py
jgieringer: yeah just in case there's an exception right at the beginning and neither LLMs are instantiated

emily-vanark:
> > however the test_pipeline_error_handling live test failed. Could we look into that before merging, please?
> 
> Fixed! [84d3708](https://github.com/SpringCare/VERA-MH/pull/126/commits/84d37082805e148da33f910e4691d689502bb9d3)

I agree!

emily-vanark (approved):

PR #137, opened 2026-04-13
Cache claude

jgieringer:
## Description
Anthropic's chat API has caching turned off by default.
With the growing chat history per turn, this PR ensures caching for claude.

## Result
Each claude model has a threshold before caching begins, but once it does, we can note it in a convo log's metadata:
```json
{
    "raw_metadata": {
        "id": "msg_01CUeyzM5CnzzSDLzZQZ7zTK",
        "model": "claude-haiku-4-5-20251001",
        "stop_reason": "end_turn",
        "stop_sequence": null, 
        "usage": {
            "cache_creation": {
                "ephemeral_1h_input_tokens": 0, 
                "ephemeral_5m_input_tokens": 64
            },
            "cache_creation_input_tokens": 64,
            "cache_read_input_tokens": 4579,
            "input_tokens": 3,
            "output_tokens": 9,
            "server_tool_use": null,
            "service_tier": "standard",
            "inference_geo": "not_available"
        }
    }, 
        "stop_details": null, 
        "model_name": "claude-haiku-4-5-20251001", 
        "model_provider": "anthropic"
}
```

### DIFF
--- a/llm_clients/claude_llm.py
+++ b/llm_clients/claude_llm.py
@@ -13,6 +13,11 @@ from .llm_interface import JudgeLLM, Role
 
 T = TypeVar("T", bound=BaseModel)
 
+# Anthropic-only: prompt caching is opt-in per request (not `cache_control` elsewhere).
+# Default ephemeral TTL is 5m (no `ttl` key). Override with ctor kwarg
+# anthropic_cache_control, e.g. {"type": "ephemeral", "ttl": "1h"}, or None to disable.
+_DEFAULT_ANTHROPIC_CACHE_CONTROL: Dict[str, Any] = {"type": "ephemeral"}
+
 
 class ClaudeLLM(JudgeLLM):
     """Claude implementation using LangChain."""
@@ -37,6 +42,9 @@ class ClaudeLLM(JudgeLLM):
     ):
         first_message = kwargs.pop("first_message", None)
         start_prompt = kwargs.pop("start_prompt", None)
+        self._anthropic_cache_control: Optional[Dict[str, Any]] = kwargs.pop(
+            "anthropic_cache_control", dict(_DEFAULT_ANTHROPIC_CACHE_CONTROL)
+        )
         super().__init__(
             name,
             role,
@@ -120,7 +128,10 @@ class ClaudeLLM(JudgeLLM):
 
         async def _invoke() -> str:
             start_time = time.time()
-            response = await self.llm.ainvoke(messages)
+            invoke_kw: Dict[str, Any] = {}
+            if self._anthropic_cache_control:
+                invoke_kw["cache_control"] = self._anthropic_cache_control
+            response = await self.llm.ainvoke(messages, **invoke_kw)
             end_time = time.time()
 
             model = (
@@ -147,6 +158,12 @@ class ClaudeLLM(JudgeLLM):
                         "total_tokens": usage.get("input_tokens", 0)
                         + usage.get("output_tokens", 0),
                     }
+                    for ck in (
+                        "cache_creation_input_tokens",
+                        "cache_read_input_tokens",
+                    ):
+                        if usage.get(ck) is not None:
+                            self._last_response_metadata["usage"][ck] = usage[ck]
                 self._last_response_metadata["stop_reason"] = metadata.get(
                     "stop_reason"
                 )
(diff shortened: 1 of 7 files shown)


llm_clients/claude_llm.py
```diff
@@ -13,6 +13,11 @@
 
 T = TypeVar("T", bound=BaseModel)
 
+# Anthropic-only: prompt caching is opt-in per request (not `cache_control` elsewhere).
+# Default ephemeral TTL is 5m (no `ttl` key). Override with ctor kwarg
```
emily-vanark: I had to google this to make sure I understood it.  Maybe add "time after which the stored context will expire and be removed from memory if it is not used"?

emily-vanark (approved):
do we need any updates on the tests for claude_llm?

llm_clients/openai_llm.py
```diff
@@ -118,7 +125,10 @@ async def generate_response(
 
         async def _invoke() -> str:
             start_time = time.time()
-            response = await self.llm.ainvoke(messages)
+            response = await self.llm.ainvoke(
+                messages,
+                prompt_cache_key=self.conversation_id,
```
luca-belli, line 130: does it mean we are also cashing OpenAI?

llm_clients/claude_llm.py
```diff
@@ -13,6 +13,12 @@
 
 T = TypeVar("T", bound=BaseModel)
 
+# Anthropic-only: prompt caching is opt-in per request (not `cache_control` elsewhere).
+# Default ephemeral TTL is 5m (no `ttl` key).
+# TTL is the time after which the stored context will expire
+# and be removed from memory if it is not used.
+_DEFAULT_ANTHROPIC_CACHE_CONTROL: Dict[str, Any] = {"type": "ephemeral"}
```
luca-belli, line 20: I think it would be nice to have this as an option in the command line that is ignored for all the other providers

llm_clients/claude_llm.py
```diff
@@ -120,7 +129,10 @@ async def generate_response(
 
         async def _invoke() -> str:
             start_time = time.time()
-            response = await self.llm.ainvoke(messages)
+            invoke_kw: Dict[str, Any] = {}
+            if self._anthropic_cache_control is not None:
+                invoke_kw["cache_control"] = self._anthropic_cache_control
+            response = await self.llm.ainvoke(messages, **invoke_kw)
```
luca-belli, line 144: to make sure i understand: we are just passing an extra arg to claude, and it's going ot be ok to keep passing the full history each turn? we don't have to do anything special?

sator-labs (approved):
LGTM, but I'd like to understand if we are also cashing OAI

llm_clients/openai_llm.py
jgieringer: openai auto caches, anthropic is the only one that doesn't. 
This added `prompt_cache_key` helps efficiently route to server w/ existing cache which can reduce cost/latency 🚀

llm_clients/claude_llm.py
jgieringer: I was thinking this would be a solid set-it-and-forget-it. Do you see any reason why this should be a command line option? Like if it isn't used (turned off), then do we try and turn off caching for the other providers?

llm_clients/claude_llm.py
jgieringer: As far as I understand, yes. We're telling Anthropic to cache the prefix as it grows and you'll see in the metadata how much cache was read as input (`"cache_read_input_tokens": 4579,` from the description)

llm_clients/claude_llm.py
luca-belli: that's a good question. in general i like to have all the option listed in one place, otherwise it's hard to hunt them down later on. i don't have an immediate use case in mind of why we would want to turn it off tbh, but i have been surprised in the past

split the difference? have a kwarg that defaults to true (e.g. caching=True) and here you just do `if caching: XXX`?
if that's too annoying for this PR, could we add a comment in the main file saying that claude is caching and that OAI is doing by default?

llm_clients/claude_llm.py
luca-belli: thanks!

PR #142, opened 2026-04-16
Ensure consistent rebuild of results in judging (especially for --resume)

jgieringer:
### Description
**Problem:** On `--resume`, `judge_conversations` only aggregated in-process results when writing `results.csv`, so a partial resumed run could overwrite the results CSV with _only_ newly finished rows and drop rows for jobs errored/skipped because their TSVs already existed.

**Change:** When the evaluation folder contains any per-job `*.tsv` files, write `results.csv` by rebuilding from all TSVs via `build_dataframe_from_tsv_files` (same path as scoring). Only writes the judging results in-memory when there are no TSVs (e.g. tests mocking the batch).

**Tests:** Added coverage for resume-with-empty batch plus an integration check that `results.csv` matches an independent TSV rebuild (using `read_csv(..., keep_default_na=False)` for empty-string parity).

**Docs:** README already described TSV rebuild on resume; behavior now matches.


emily-vanark:
Question:  If we're resuming, does that mean no results.csv was written before?  Or just that we're filling in gaps?  Or maybe both are possible?

I guess I wonder if it's worth looking to see if there's an existing results.csv and then concatenating it, or if we're confident that the rebuild isn't loosing us any info, so we just want to do that.  (I've been bitten in the past when we added columns to the OG results.csv, but I didn't realize we wanted to reconstruct those in the rebuild for.... a while.)

judge/runner.py
```diff
@@ -564,24 +565,37 @@ async def judge_conversations(
...
-            if k
-            not in ["filename", "run_id", "judge_model", "judge_instance", "judge_id"]
-        ]
-        pd.DataFrame(results, columns=cast(Any, columns)).to_csv(
-            f"{output_folder}/{filename}", index=False
-        )
+    if save_aggregated_results:
+        csv_name = filename or "results.csv"
+        out_csv = os.path.join(output_folder, csv_name)
+        has_eval_tsvs = bool(_index_existing_evaluation_tsv_basenames(output_folder))
```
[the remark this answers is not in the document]
jgieringer: Fixed this a better way

jgieringer:
> Question: If we're resuming, does that mean no results.csv was written before? Or just that we're filling in gaps? Or maybe both are possible?

If first time judging finishes, but skipped some evals due to error, then a results.csv is written! So the bug was that on resuming, initial results were at risk of being overwritten as the process was only writing the latest (which were the initially skipped!)

> I guess I wonder if it's worth looking to see if there's an existing results.csv and then concatenating it, or if we're confident that the rebuild isn't loosing us any info, so we just want to do that. (I've been bitten in the past when we added columns to the OG results.csv, but I didn't realize we wanted to reconstruct those in the rebuild for.... a while.)

@emily-vanark this is a great question. Cursor added a test to ensure rebuilding doesn't lose any info. But we've seen it happen, but I believe that's fixed in #121?

I added test `test_rebuilt_dataframe_matches_committed_results_csv` that asserts the initial (real-world-built!) results.csv == the rebuilt-from-TSVs results.csv and it's passing!

tests/integration/test_evaluation_runner.py
```diff
@@ -480,6 +481,62 @@ async def test_judge_conversations_basic_workflow(
         assert "judge_model" in df.columns
         assert "judge_instance" in df.columns
 
+    async def test_results_csv_matches_tsv_reconstruction(
+        self,
```
emily-vanark, line 485: love this as a concept.... think we might need to make it an integration test that uses real judging results to make it meaningful.

jgieringer: I know.. it's tough bc this results.csv is from a real run and from real TSVs, but if we update anything, it will still be comparing to old fixture data..
Idea: let's never change the rubric 🥇

emily-vanark: Could we integrate this into one of the live integration tests we already have?  Maybe the pipeline one?  Before the tmp results.csv gets lost, rename it, then re-construct it from the tmp tsv's and make sure the two match?

emily-vanark: (The risk is less from the rubric changing and more from us adding or removing fields from the main results.csv and the reconstruction doesn't keep up with... I think... although I guess rubric changes that changed the dimensions could also break it?)

emily-vanark (approved):
Would like to spend a few minutes considering whether we could add an integration test for the rebuild pretty easily, but would rather have this in than out, so if that seems like a bigger time suck, I'm okay to merge as-is.

tests/integration/test_evaluation_runner.py
jgieringer: Since rebuilding the results from TSVs looks through the loaded DIMENSIONS from the rubric, any code changes that end up with a different fields,etc.  I believe we're safe and this unit test will fail and call for an update to this test & its fixture. I think that's safe for now!

judge/runner.py
```diff
@@ -564,24 +565,44 @@ async def judge_conversations(
...
-            if k
-            not in ["filename", "run_id", "judge_model", "judge_instance", "judge_id"]
-        ]
-        pd.DataFrame(results, columns=cast(Any, columns)).to_csv(
-            f"{output_folder}/{filename}", index=False
-        )
+    if save_aggregated_results:
+        csv_name = filename or "results.csv"
+        out_csv = os.path.join(output_folder, csv_name)
+        # On resume, skip a second listdir if pre-batch TSVs existed; otherwise
```
luca-belli: could you explain this a little more? I am confused by the comment

sator-labs (approved):

PR #143, opened 2026-04-20
Adding coverage for multi-session runner and LLMInterface hooks

luca-belli:
## Summary

A set of improvements to the conversation runner and LLM interface, developed while integrating a new provider type internally. All changes are generic and provider-agnostic.

Note: the custom LLM class will need to have knowledge of session requirements (e.g., an intake session is always mandated).

**Multi-session support** (`runner.py`, `generate.py`)
- The runner now executes an ordered list of session types per conversation (via `--sessions`). A single session without `--sessions` follows the existing code path unchanged.
- `agent.setup()` is called once before the session loop; `agent.first_speaker` is respected per session, overriding the global `--provider-speaks-first` flag when set.
- New extensibility hooks on `LLMInterface`: `prepare_sessions()`, `finish_and_reset_session()`, `first_speaker`, `setup()`.

**Bespoke termination signals** (`llm_interface.py`, `conversation_simulator.py`)
- Providers can declare `custom_termination_signals` to end a conversation early on provider-specific strings (e.g. error tags).
- `_extract_signals()` detects signals from the raw response before `_post_process_response()` strips artifacts.
- `_should_discard_response()` lets a provider remove an error turn from the transcript entirely rather than marking it as early termination.

**Debug output** (`conversation_simulator.py`)
- Turn-by-turn progress is printed via `debug_print` (only active with `--debug`).
- Raw response is recorded in turn metadata for the provider side.

## Test plan

- [ ] `python3 generate.py -u claude-sonnet-4-6 -p claude-sonnet-4-6 -t 6 -r 1` — single session, existing behaviour unchanged
- [ ] `python3 generate.py ... --sessions session_a,session_b` — two sessions run in sequence, filenames suffixed
- [ ] `python3 generate.py ... --debug` — turn-by-turn output visible


sator-labs:
Important note: it currently fails on one (1) unit test

tests/integration/test_conversation_runner.py
```diff
@@ -335,15 +335,17 @@ async def test_run_single_conversation_creates_files(
         assert result["llm1_model"] == "mock-persona-model"
         assert result["llm1_prompt"] == "TestPersona"
         assert result["run_number"] == 1
-        assert result["turns"] == 4
-        assert "filename" in result
-        assert "log_file" in result
+        assert result["turns"] == 5
```
luca-belli: note: I need to understand why this fix is necessary

jgieringer:
Since we're so close to shipping v1.1, can I trouble you @sator-labs to change merging from main to v1.1?
That way we'll know what conflicts arise and can solve them in this branch vs. merging to main & having to solve conflicts up the chain?

sator-labs:
> Since we're so close to shipping v1.1, can I trouble you @sator-labs to change merging from main to v1.1? That way we'll know what conflicts arise and can solve them in this branch vs. merging to main & having to solve conflicts up the chain?

done

generate_conversations/runner.py
```diff
@@ -222,30 +224,37 @@ async def run_single_conversation(
...
     ) -> Dict[str, Any]:
-        """Run a single simulated conversation (persona vs provider LLM).
+        """Run a simulated conversation (persona vs provider LLM),
+          across one or more sessions.
 
-        Uses fresh LLM instances per call; safe for concurrent use. Logs turns,
-        writes transcript to self.folder_name, then cleans up logger and LLMs.
+        A single session without --sessions is treated as a one-element session list so
+        the same code path handles both cases. The agent's prepare_sessions() hook
+        normalises the list (e.g. prepending INTAKE for ray-backend).
```
jgieringer: Can a non-internal example be provided or is this truly spring-related?

generate_conversations/runner.py
```diff
@@ -293,78 +303,136 @@ async def run_single_conversation(
...
+                    log_conversation_end(
+                        logger=session_logger,
+                        total_turns=len(conversation),
+                        early_termination=session_early_term,
+                        total_time=session_time,
+                    )
+
+                    output_filename = f"{session_filename_base}.txt"
+                    simulator.save_conversation(output_filename, self.folder_name)
+                    all_conversations.extend(conversation)
```
jgieringer: I'm finding extending conversation, a list of turns, into list all_conversations is confusing.
Could you rename it to be all_turns or full_conversation?

generate_conversations/runner.py
```diff
@@ -293,78 +303,136 @@ async def run_single_conversation(
...
+                    filenames.append(f"{self.folder_name}/{output_filename}")
+                    log_files.append(
+                        f"logging/{self.run_id}/{session_filename_base}.log"
+                    )
```
jgieringer: Making note I will need to update https://github.com/SpringCare/VERA-MH/pull/144 to store logging location

jgieringer (changes requested):
If sessions could be desired by a non-spring entity, could you provide detail in the README (or its own md file) on exactly what it is doing?
It's multiple short "conversations" that get stitched together?
So maybe even all_conversations should be something like full_conversation?

generate_conversations/conversation_simulator.py
```diff
@@ -23,18 +24,27 @@ def __init__(self, persona: LLMInterface, agent: LLMInterface):
...
+        response: str,
+        speaker: LLMInterface,
+        extracted_signals: Optional[List[str]] = None,
     ) -> bool:
+        """Check if the response indicates the conversation should end.
+
+        Two termination paths:
+        - Global signal (self.termination_signal): only the persona can trigger it,
+          checked against the post-processed response.
+        - Bespoke signals: passed in as extracted_signals, detected from the raw
```
emily-vanark: Will the term "bespoke" make sense to all our users?  I think I'd go for "custom for specific LLM client" if I'm understanding correctly what it is?   (More curiosity than blocking.)

generate_conversations/conversation_simulator.py
```diff
@@ -23,18 +24,27 @@ def __init__(self, persona: LLMInterface, agent: LLMInterface):
...
+        # Global signal: persona-only, checked on cleaned response
+        if speaker == self.persona and re.search(
+            re.escape(self.termination_signal), response, re.IGNORECASE
+        ):
+            return True
 
-        # Check for exact phrase matches (case insensitive)
-        if re.search(re.escape(self.termination_signal), response, re.IGNORECASE):
+        # Bespoke signals extracted from raw response before post-processing
+        if extracted_signals and any([signal in speaker.bespoke_termination_signals for signal in extracted_signals]):
```
emily-vanark: this looks like it's looking for an exact match... for the global signals we do a regex search with an ignorecase... do we want to offer the same for the bespoke signals?

generate_conversations/conversation_simulator.py
```diff
@@ -66,8 +76,21 @@ async def generate_conversation(
             current_speaker = self.agent
             next_speaker = self.persona
 
+        debug_print(
+            f"[SIM] Starting conversation: persona_speaks_first={persona_speaks_first}, "
```
emily-vanark: What's the "[SIM]" indicating?

emily-vanark: Also.... how is the debug_print turned on or off?

emily-vanark: I guess I'm wondering if we want all these debug_prints in the code base longer term.

emily-vanark: (And if we're keeping them, we should add to the README how to turn them on/ off? and maybe use that pattern more places?)

generate_conversations/runner.py
emily-vanark: I... am not clear what ray-backend is, or why I'd want to prepend INTAKE to it.  This could use more explanation... maybe it's trying to say, "For example, if an agent has session types INTAKE and RETURN, each type might be added to a persona name, e.g. Ray, bcoming INTAKE-Ray and RETURN-Ray."?   Some sort of explanation like this would help me.

generate_conversations/runner.py
```diff
@@ -222,30 +224,37 @@ async def run_single_conversation(
...
-        writes transcript to self.folder_name, then cleans up logger and LLMs.
+        A single session without --sessions is treated as a one-element session list so
+        the same code path handles both cases. The agent's prepare_sessions() hook
+        normalises the list (e.g. prepending INTAKE for ray-backend).
 
         Args:
             persona_config (dict): Must have "prompt" and "name". Persona LLM
                 identity comes from ``self.persona_model_config`` (including ``model``).
-            max_turns (int): Max conversation turns for a conversation.
+            max_turns (int): Max conversation turns per session.
```
emily-vanark, line 343: I'm trying to wrap my head around whether we'd want max_turns to be the max_turns per session or the max_turns across sessions.  It feels like if the conversations are going to be concatenated together for judging, we'd want it to be across sessions for consistency with single-session evalautions.  But if the sessions are going to be evaluated as independent conversations, then we'd want max_turns to apply on the session level...

generate_conversations/runner.py
```diff
@@ -222,30 +224,37 @@ async def run_single_conversation(
...
             conversation_index (int): Index in the batch of conversations.
             run_number (int): Run index for this prompt (e.g. 1 of runs_per_prompt).
             **kwargs: Unused; reserved for future use.
 
         Returns:
             Dict[str, Any]: index, llm1_model, llm1_prompt, run_number, turns,
-            filename, log_file, duration, early_termination, conversation.
+            filenames, log_files, duration, early_termination, conversation, and
+            optional filename/log_file for single-transcript compatibility.
```
emily-vanark: I am confused... what would be the difference between log_files and filename/log_file... and... is this taking into account the udpates coming in [!144](https://github.com/SpringCare/VERA-MH/pull/144) ?

generate_conversations/runner.py
```diff
@@ -222,30 +224,37 @@ async def run_single_conversation(
...
         persona_name = persona_config["name"]
 
-        # Generate filename base using persona name, model, and run number
         tag = uuid.uuid4().hex[:6]
-        filename_base = f"{tag}_{persona_name}_{model_name}_run{run_number}"
+
+        workflow = self.agent_model_config.get("workflow", "")
+        workflow_part = f"_{workflow}" if workflow else ""
+        filename_base = (
+            f"{tag}_{persona_name}_{model_name}{workflow_part}_run{run_number}"
```
emily-vanark, line 363: I wonder if introducing this naming scheme will break anything in our utils for parsing file names / need updates there to handle this approach?

generate_conversations/runner.py
emily-vanark: It's true that I assumed all_conversation was a list of conversation objects... if we're extending it so all_converastions is a single, longer-conversation from concatenated sessions, I agree with @jgieringer 's naming suggestions.  (And I feel like that leans towards setting max_turns over all the sessions rather than per session.)

llm_clients/llm_interface.py
```diff
@@ -311,6 +311,77 @@ def _post_process_response(self, response: str) -> str:
         """
         return response
 
+    @property
+    def bespoke_termination_signals(self) -> List[str]:
+        """Per-implementation early-termination signals.
+
+        Return a list of signal strings. When any signal is found in this
+        speaker's response, the conversation ends early.
```
emily-vanark, line 340: Is it for both the agent as user and the agent as provider?  I've been assuming only as provider, but this makes it sound like it could be either....

generate_conversations/conversation_simulator.py
emily-vanark: Is this being called after ` _extract_signals` in the llm_interface has already handled that?

tests/unit/generate_conversations/test_conversation_simulator.py
```diff
@@ -503,3 +503,133 @@ async def test_provider_first_first_message(self):
...
+        assert history[-1]["speaker"] == "provider"
+        assert history[-1]["early_termination"] is True
+
+    async def test_discard_response_removes_turn_from_history(self):
+        """When _should_discard_response is True, the offending turn is dropped."""
+
+        class ErrorAgent(MockLLM):
+            @property
+            def bespoke_termination_signals(self):
+                return ["[ERROR]"]
```
emily-vanark, line 540: I find myself wondering if / how this interacts with the retry_on_error additions Josh put in... maybe they're independent?

emily-vanark:
> If sessions could be desired by a non-spring entity, could you provide detail in the README (or its own md file) on exactly what it is doing? It's multiple short "conversations" that get stitched together? So maybe even all_conversations should be something like full_conversation?



> If sessions could be desired by a non-spring entity, could you provide detail in the README (or its own md file) on exactly what it is doing? It's multiple short "conversations" that get stitched together? So maybe even all_conversations should be something like full_conversation?

Agree we need to update the README a couple places:
* probably a note in the section about setting up your own llm_interface
* also adding it to the parameters documentation (an example would be great for formatting... is it `[INTAKE, RETURN]` or `INTAKE, RETURN` or `INTAKE RETURN` or... something with quotes?
(This interfaces with [!145](https://github.com/SpringCare/VERA-MH/pull/145) as well.)  

 ... and... did we get the argument for sessions in both generate.py and the pipeline?  I feel like maybe we missed the pipeline... and also it'll belong (I think) in [this](https://github.com/SpringCare/VERA-MH/pull/145) once it merges...

emily-vanark (commented):
In general, I think this is a fine and good addition to the repo.  Also, I'm a bit concerned at its overlap with several other open PRs right now... maybe we want to let those finalize in the next day and then mix this in with the rest?

generate_conversations/runner.py
luca-belli: each session is independent, so the max would be num_session*max

i think in an ideal case there should be per session max (e.g. session 1 has max of n1, session 2 has max of n2 etc)

generate_conversations/runner.py
luca-belli: that might be a bug, because they are separate. let me check

llm_clients/llm_interface.py
luca-belli: good question
it was meant only for provider, but in principle there is nothing that prevents the user from implementing that too
the other option would be to separate User and Provider classes, but it might be too big of a refactor now.

I'll add a doc string to clarify

generate_conversations/runner.py
jgieringer: Was this a bug?

README.md
```diff
@@ -200,8 +200,11 @@ uv run python run_pipeline.py --help
...
+| | `--sessions` | Comma-separated session types to run in order for each persona/run (e.g. `intake,coaching`). Omit for a single session (default). |
 | | `--resume` | Continue a previous run: set `--output` to the existing `p_*` run directory; skips persona/run pairs that already have transcript files. User/provider models, turns, and runs must match the original run (see `generate.py` validation). |
 
+**Multi-session conversations:** Use `--sessions` when the provider under test supports distinct session types (for example intake followed by coaching). For each persona and run, VERA-MH runs those sessions sequentially—each gets its own transcript and log, with filenames like `..._run1_1_intake.txt` and `..._run1_2_coaching.txt`. The provider LLM can implement `prepare_sessions()`, `enter_session()`, and `first_speaker` on [`LLMInterface`](llm_clients/llm_interface.py) to set up API sessions, enforce prerequisites, or control who speaks first per session type (see [Connecting your own LLM, Agent, or API](#connecting-your-own-llm-or-api)). With `--resume`, already-finished sessions are skipped individually so you can continue a partial multi-session run.
```
jgieringer: > `..._run1_1_intake.txt` and `..._run1_2_coaching.txt`

nit: add an s in front of session count `..._run1_s1_intake.txt` and `...run1_s2_coaching.txt`

jgieringer (approved):
Tests all passed. Just one nit pick!

PR #151, opened 2026-04-24
look for p_ convo output

jgieringer:
## Description
With the new output defaults, conversations are now stored under p_.../conversations/*.txt instead of conversations/p_.../*.txt

When passing p_.../conversations as the input to judge, it creates a judge folder called j_...conversations instead of j_...p_... which caused "Unknown" User name and Agent name 💀 

This PR addresses looking one level up for the expected p_... dir name to preserve user and agent information

### DIFF
--- a/utils/conversation_layout.py
+++ b/utils/conversation_layout.py
@@ -16,11 +16,13 @@ def resolve_conversation_input(folder: str) -> tuple[str, Optional[str], str]:
 
     - If ``folder/conversations/`` contains ``.txt`` files, ``folder`` is treated as
       a generation run root; transcripts are read from ``folder/conversations``.
-    - If ``folder`` is named ``conversations``, contains ``.txt`` files, and its
-      parent's basename passes ``is_generation_run_folder_basename`` (canonical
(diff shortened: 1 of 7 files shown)


emily-vanark:
Hmm... running `uv run pytest -m "not live"` on this branch still results in the mysterious adhoc outputs...   (I guess I was wrong about it being the live tests? Updated the issue.)

Stll getting 69 judge_unscoped files similar to the ones in the issue and an empty single_datetime_conv folder.

utils/conversation_layout.py
emily-vanark, line 36: Not necessarily with this PR, but... I was refreshing myself on what os.path.basename does and it seems like using Pathlib might be a more modern / robust way of doing this (doesn't return an empty string if there's a trailing '/' at the end of the folder name).   Maybe we're already passing in pathlib objects so this will be okay?

emily-vanark: (I guess the normpath actually handles this so... definitely not important today.)

emily-vanark (approved):
Tests pass.  LGTM.

PR #157, opened 2026-05-28
Log Failures + Use "json_schema" for Gemini Structured Outputs

jgieringer:
## Summary

Fixes intermittent Gemini judge failures (`Response is not an instance of QuestionResponse`) by switching Gemini structured output to native **`json_schema`** mode and adding structured-output failure diagnostics across the judge pipeline.

## Problem

When judging with `gemini-3.5-flash` (and related Gemini models), roughly **20% of conversations** failed with:

```text
LLM call failed after 4 attempts: Response is not an instance of QuestionResponse
```

Root cause: LangChain’s default Gemini path uses **`method="function_calling"`** — the Pydantic `QuestionResponse` model is exposed as a **tool**, and the response is parsed via `PydanticToolsParser`. When Gemini returns no tool call (plain text, partial response, safety block, etc.), the parser yields **`None`**, which fails our `isinstance` check after retries.

Per-conversation logs often stopped at the last `--- Question N ---` with **no** `STRUCTURED RESPONSE`, so there was no visibility into what the model actually returned.

**Note:** Using Pydantic (`QuestionResponse`) does not automatically use Gemini JSON schema mode. Pydantic defines the *shape*; LangChain still chooses *how* to ask Gemini for that shape (`function_calling` vs `json_schema`).

### 1. Use `json_schema` for Gemini structured judging (`llm_clients/gemini_llm.py`)

- `with_structured_output(response_model, method="json_schema", include_raw=True)`
- Parse via LangChain’s `include_raw` payload (`parsed`, `raw`, `parsing_error`)
- Coerce dict → Pydantic when needed; richer `ValueError` on failure
- Attach `structured_output_debug` to `last_response_metadata` on retry/failure

### 2. Surface failures in judge logs and batch output

- **`judge/llm_judge.py`**: On `LLMGenerationFailed`, log `STRUCTURED OUTPUT FAILED for question {id}`, `structured_output_debug`, and retry metadata to the per-conversation log under `{run}/logs/`.
- **`judge/runner.py`**: Print full `last_response_metadata` JSON when a conversation evaluation fails.
- **`judge.py`**: Add `--debug` / `-d` (aligned with `generate.py`) for stdout debug output from `gemini_llm`.

### 3. Tests

- Update `tests/unit/llm_clients/test_gemini_llm.py` for `json_schema` + `include_raw` mocks and a test asserting failure debug metadata.

## Why `json_schema` for Gemini (not just Pydantic)

| | `function_calling` (previous default) | `json_schema` (this PR) |
|---|---|---|
| **API contract** | Model must emit a **tool call** with args matching the schema | Model returns **`application/json`** constrained by schema from the same Pydantic model |
| **Fit for judge output** | Tool framing is heavier than needed for a simple `{answer, reasoning}` object | Direct JSON object matches judge rubric responses |
| **Failure mode** | Missing/malformed tool calls → `parsed is None` → opaque `QuestionResponse` error | Same schema source, but Gemini’s native structured output is typically more stable for flat JSON |
| **Pydantic role** | Generates tool schema | Generates `response_json_schema` via `model_json_schema()` — still validated into `QuestionResponse` in Python |

We still use **`QuestionResponse`** end-to-end; this PR changes the **transport** LangChain uses with Gemini, not the rubric contract.

## Files changed

- `llm_clients/gemini_llm.py` — `json_schema`, raw capture, failure diagnostics
- `judge/llm_judge.py` — log structured-output failures per question
- `judge/runner.py` — print LLM metadata on conversation failure
- `judge.py` — `--debug` flag
- `tests/unit/llm_clients/test_gemini_llm.py` — updated/new tests


judge/runner.py
```diff
@@ -86,13 +105,17 @@ async def _evaluate_single_conversation_with_judge(
...
+    try:
+        evaluation = await judge.evaluate_conversation_question_flow(
+            conversation,
+            output_folder=output_folder,
+            auto_save=True,
+            verbose=False,
+            judge_instance=judge_instance,
+        )
+    except LLMGenerationFailed:
+        _print_llm_failure_metadata(judge, conversation_filename, judge_model)
```
luca-belli: i don't think we should print on screen by default. we should log and print if debug?

llm_clients/gemini_llm.py
```diff
@@ -13,6 +14,74 @@
...
+def _truncate_for_log(
+    value: Any, max_chars: int = _STRUCTURED_OUTPUT_MAX_LOG_CHARS
+) -> str:
+    text = value if isinstance(value, str) else repr(value)
+    if len(text) <= max_chars:
+        return text
+    return f"{text[:max_chars]}... [truncated, {len(text)} chars total]"
+
+
+def _raw_message_content_for_log(raw: Any) -> str:
```
luca-belli, line 34: two thoughts:

1. wouldn't be smoother to just change the function to 


```
content = XXX

return content[:MAX]
```

so we don't need the aboce

2. don't we want the log to have all the excruciating details? isn't that the point of the log?

llm_clients/gemini_llm.py
```diff
@@ -13,6 +14,74 @@
...
+    """Bounded text preview of the raw model message body."""
+    if raw is None:
+        return ""
+    if isinstance(raw, AIMessage):
+        return _truncate_for_log(raw.content)
+    return _truncate_for_log(raw)
+
+
+def _serialize_raw_message(raw: Any) -> Dict[str, Any]:
```
luca-belli, line 43: could you add an extra sentence in the doc?

i understand it simplifies the data, but also later is says it omits the content, so i am confused

llm_clients/gemini_llm.py
```diff
@@ -13,6 +14,74 @@
...
+    if isinstance(invoke_result, dict) and "parsed" in invoke_result:
+        return (
+            invoke_result.get("parsed"),
+            invoke_result.get("raw"),
+            invoke_result.get("parsing_error"),
+        )
+    return invoke_result, None, None
+
+
+def _coerce_structured_response(response: Any, response_model: Type[T]) -> Optional[T]:
```
luca-belli: is that just a gemini thing or should this me more general?

llm_clients/gemini_llm.py
```diff
@@ -13,6 +14,74 @@
...
+        if raw.additional_kwargs:
+            payload["additional_kwargs"] = dict(raw.additional_kwargs)
+        return payload
+    return {"type": type(raw).__name__, "repr": _truncate_for_log(raw)}
+
+
+def _extract_include_raw_result(
+    invoke_result: Any,
+) -> tuple[Any, Any, Any]:
+    """Unpack LangChain include_raw=True structured output payloads."""
```
luca-belli: could you add a simple toy example? I am not sure i am able to parse it without knowing more of the context

llm_clients/gemini_llm.py
```diff
@@ -174,11 +243,36 @@ async def _invoke() -> str:
 
         return await self._run_with_retry(_invoke, provider="gemini")
 
+    def _build_structured_output_failure_debug(
+        self,
+        *,
```
luca-belli, line 235: do we need this?

llm_clients/gemini_llm.py
```diff
@@ -193,27 +287,72 @@ async def generate_structured_response(
...
+            raise ValueError(
+                f"Expected {response_model.__name__}, got {type(parsed).__name__}: "
+                f"{_truncate_for_log(parsed)}. "
+                f"raw_content={failure_debug.get('raw_content')!r}"
+            )
 
-            return response  # type: ignore[return-value]
-
-        return await self._run_with_retry(_invoke, provider="gemini")
+        def _on_structured_error(
```
luca-belli, line 329: i know we have error handling somewhere in the main loop

how does it interact with it?

judge/runner.py
```diff
@@ -25,6 +28,22 @@
 EVALUATION_SEPARATOR = ":"
 
 
+def _print_llm_failure_metadata(
```
luca-belli: this function should be in utils, not in the main runner

llm_clients/gemini_llm.py
jgieringer: `_on_structured_error` feeds judge log files (via last_response_metadata), not stdout, so the runner only sees the final `LLMGenerationFailed` after all retries are exhausted

Gemini got the extra hook because its `json_schema` + `include_raw` path often succeeds at the API but returns `parsed=None` with useful info in `raw` / `parsing_error`, so worth preserving for judge logs.

llm_clients/gemini_llm.py
jgieringer: yeah the two helper situation only for gemini is clunky.
I was fine with it because it's extra protection, but it might not be needed. Thinking.

PR #159, opened 2026-06-03
Isolate Human-validated Conversations

jgieringer:
## Description
publication_data/human_validation/conversations contained more conversations than were analyzed.

This PR adds dir publication_data/human_validation/conversations_analyzed to isolate the 90 reviewed conversations


luca-belli (commented):
double checking we only have partial information on each run?
could we have a readme (one line is fine) that says those are the conversation for XXX paper, with a public link to it?

luca-belli (approved):

PR #160, opened 2026-06-04
support removing deprecated params for LLMs

jgieringer:
## Description
Opus 4.8 was released and [deprecated params](https://platform.claude.com/docs/en/about-claude/model-deprecations#api-parameter-deprecations) that were set by default by judge init.

This PR filters deprecated params & adds temperature filter for claude opus 4.8

### DIFF
--- a/llm_clients/claude_llm.py
+++ b/llm_clients/claude_llm.py
@@ -1,5 +1,5 @@
 import time
-from typing import Any, Dict, List, Optional, Type, TypeVar
+from typing import Any, ClassVar, Dict, List, Optional, Type, TypeVar
 
 from langchain_anthropic import ChatAnthropic
 from langchain_core.messages import HumanMessage, SystemMessage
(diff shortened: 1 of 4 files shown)


llm_clients/claude_llm.py
```diff
@@ -27,6 +27,10 @@ class ClaudeLLM(JudgeLLM):
     ``anthropic_cache_control`` constructor args).
     """
 
+    _UNSUPPORTED_MODEL_PARAMS: ClassVar[Dict[str, frozenset[str]]] = {
```
luca-belli: i like this idea a lot!
why are you using a var? wouldn't be more consistent with the rest to have a function that returns unsupported parameters instead?

luca-belli (approved):
so I assume that the idea is that every time we find a new non supported parameter, we add it to this?

llm_clients/claude_llm.py
jgieringer: yesss thank you - I let the AI do AI things

PR #161, opened 2026-06-04
Update recommended run

jgieringer:
## Description
GPT 5.4 has been prescribed by clinicians as latest recommended judge

This PR:
- fixes bug to load scored results.csv files correctly
- adds v1.1.1 in score assets
- refactors recommended script to judge with GPT 5.4

### DIFF
--- a/judge/score_utils.py
+++ b/judge/score_utils.py
@@ -664,7 +664,7 @@ def has_dimension_data(df: pd.DataFrame) -> bool:
     return any(dim in df.columns and df[dim].notna().any() for dim in DIMENSIONS)
 
 
-def read_results_csv(results_csv_path: Union[str, Path]) -> pd.DataFrame:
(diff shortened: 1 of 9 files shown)


judge/score_utils.py
```diff
@@ -664,6 +664,16 @@ def has_dimension_data(df: pd.DataFrame) -> bool:
     return any(dim in df.columns and df[dim].notna().any() for dim in DIMENSIONS)
 
 
+def read_results_csv(results_csv_path: Union[str, Path]) -> pd.DataFrame:
```
luca-belli: i feel this is borderline too close to an anti-pattern of redefining default names

if you think it's helpful to have, maybe use a more descriptive name like `read_judge_results_csv`, to make sure we don't use somewhere else by mistake?

luca-belli (approved):
LGTM, but i made a note for us to rethink the structure a bit going forward

PR #170, opened 2026-07-09
VERA 2.0 Architecture

luca-belli:
## Summary

Defines VERA-MH's **target architecture** — the shape the codebase migrates to as `generate.py`/`judge.py`/`run_pipeline.py` are replaced by a single `vera.py` CLI — and locks it down in two documents plus a 6-phase (+2 orthogonal) migration plan:

- **[`docs/architecture.md`](docs/architecture.md)** — the readable version: system overview, layer model, CLI surface, package responsibilities, data flow, and a phased migration plan with per-phase "Done when" bars.
- **[`docs/ARCHITECTURE-SPINE.md`](docs/ARCHITECTURE-SPINE.md)** — the terse, numbered version: 26 `AD-n` invariants, each stating what it binds, what it prevents, and the adopted rule. This is what a PR review or an agent cites (`AD-12`, not a paragraph) when a change touches a rule.
- **[`docs/vera-cli-use-cases.md`](docs/vera-cli-use-cases.md)** — the CLI/config surface in full: entity vocabulary (`u`/`c`/`j`), naming scheme, config mechanics, six worked use cases.
- **`.github/CODEOWNERS`** — enforces required review on the handful of files this doc calls "stable interfaces."

No production code changes in this PR — it's the design contract the migration will be judged against.

## Why two architecture docs instead of one

`architecture.md` and `ARCHITECTURE-SPINE.md` aren't duplicates — they serve different readers and decay at different rates:

| | `architecture.md` | `ARCHITECTURE-SPINE.md` |
|---|---|---|
| Audience | Humans getting oriented, PR descriptions, onboarding | Agents and reviewers checking one specific rule |
| Shape | Prose, tables, diagrams, migration narrative | Numbered `AD-n` entries: binds / prevents / rule |
| Changes when | Anything about the system's story changes | Rarely — spine entries are meant to be cited by ID and stay stable |
| How you use it | Read top to bottom once | Grep for `AD-10`, read one paragraph |

The spine is derived from the prose doc (`sources:` front-matter points back at it), not written independently — so the two can't silently diverge into two different "sources of truth." Every rule that shows up as prose in `architecture.md` also exists as a citable `AD-n` in the spine; the spine is what makes "which invariant does this PR violate" answerable in one line instead of a paragraph search.

## Why this justifies its own PR (and its own doc, not just a README section)

This isn't incremental documentation — it's the reason the migration is safe to do at all with heavy agent involvement:

- **The codebase is explicitly agent-coding-optimized.** Most files are fair game for an agent to change freely inside its own package's boundaries. A small, named set of **stable interfaces** (`llm_interface.py`, `queue.py`, `role.py`, `naming.py`, `config_schema.py`, `storage_backend.py`) require a design doc before modification — called out individually in CODEOWNERS, not just covered by a package-level rule, so their significance is visible at a glance instead of being one file among many in a directory.
- **Invariants are enforceable, not aspirational.** Import boundaries (`generate/` ⊥ `judge/` ⊥ `score/`, `workers/` never importing domain, `utils/` as a true leaf) are meant to be checked by import-linter + grimp in CI, phased in incrementally as each boundary is created — not just prose an agent might rationalize around.
- **ESCALATE is a hard stop, and documenting a phase doesn't pre-clear it.** Changing a stable interface, adding a subcommand, or touching import boundaries requires maintainer sign-off *at the moment of the change*, even if this doc already described that exact phase. This is the one place a human is guaranteed to be in the loop on the highest-risk edits (e.g. Phase 5's `workers/` rewrite).
- **The migration itself needed a plan a reviewer could hold an agent to.** Each of the 6 phases (0, 1–5) plus 2 orthogonal phases (S: storage abstraction, O: OpenSpec adoption) has a concrete "Done when" bar — not "phase looks done," but a specific, checkable condition (e.g. Phase 4: a 2+-rubric config produces separated output *and* a pre-existing single-rubric config still behaves identically).

## Key design decisions worth knowing about

- **Two independent pipelines, one CLI.** Generation (`u`↔`c` LLM conversation) and judging (`j` walks rubric questions) never import each other; `vera.py` is a thin orchestrator with zero business logic.
- **Interfaces live with their implementations, not in a shared `interfaces/` package.** `llm_clients/` owns `LLMInterface` + every provider, `workers/` owns `QueueProtocol` + queue implementations, `storage/` owns `StorageBackend` + backends — grouped by concern (Common Closure Principle), so an LLM-provider change and a storage-backend change never force an unrelated shared package to move.
- **Rubrics are self-describing bundles**, not a bare `.tsv` path with assumed sibling filenames — `--rubric`/`judging.rubrics[]` point at a manifest describing rubric file, prompts, and (informational-only) intended personas.
- **Traceability is a hash discipline, not a convention.** `config.json` (immutable) + `config.json.sha256` (sidecar) + `state.json` (mutable, resume state) — the sha256 is computed exactly once and reused everywhere it's needed, never independently recomputed.
- **Rubric navigation lives in code, never in the prompt.** `QuestionNavigator` deterministically walks the rubric's flow data; the judge LLM answers the current question only and never decides what comes next — this keeps evaluation deterministic and testable.
- **Data stays out of code.** Rubric/persona content must live in `data/`, never as Python constants — a non-developer must be able to add or edit a rubric without touching code.

## What's *not* in this PR

No implementation. `generate.py`/`judge.py`/`run_pipeline.py` are untouched; Phase 0 (rubric bundle manifest helper) and beyond ship as separate, individually-escalated PRs.


docs/architecture.md
```diff
@@ -0,0 +1,207 @@
+# VERA-MH Architecture
+
+Validation of Ethical and Responsible AI in Mental Health: simulate mental-health conversations, evaluate them against a clinical rubric, and aggregate scores for comparison across provider agents.
```
emily-vanark: maybe "one or more clinical rubrics"?

emily-vanark: And "aggregate scores for each rubric for comparioson..."?

docs/architecture.md
```diff
@@ -0,0 +1,207 @@
...
+                                        │
+                                        ▼
+                                   pool ──► j_pooled__*/
+```
+
+## Domain model
+
+| Concept | Location | Notes |
+|---------|----------|-------|
+| Persona | `data/personas.tsv` | Simulated patient; drives the user-side LLM |
```
emily-vanark: would probably use "User" as more neutral than "patient".

docs/architecture.md
```diff
@@ -0,0 +1,207 @@
...
+| Transcript | `p_*/conversations/*.txt` | Turn-by-turn chat log |
+| Rubric | `data/rubric.tsv` | Question flow, dimensions, severity |
+| Evaluation run | `j_*/` | TSV results, logs, metadata |
```
emily-vanark: Are we moving away from [this](https://github.com/SpringCare/VERA-MH/pull/170/changes#r3562642455) structure?  Because... that seemed logical to me.... grouping the conversations together with their evaluations...

docs/architecture.md
```diff
@@ -0,0 +1,207 @@
...
+Infrastructure (llm_clients/)
+    ↓ use
+Shared utilities (utils/)
+```
+
+**Import rules:**
+
+- Domain packages do not import each other.
+- Infrastructure does not import domain packages.
+- `utils/` is a leaf layer — it does not import domain or infrastructure packages.
```
emily-vanark: What about data files?  (I'm thinking about deriving dimension names and rating values from rubrics, which we might need in both judge and score...)

docs/architecture.md
```diff
@@ -0,0 +1,207 @@
...
+
+| Path | Role |
+|------|------|
+| `data/` | Committed evaluation inputs (personas, rubrics, prompts) |
+| `output/` | Runtime artifacts (gitignored) |
+| `scripts/` | Pipeline helpers until absorbed into `vera pool` |
+| `tests/` | Permanent tests |
+| `tmp_tests/` | Scratch experiments (not committed) |
+
+## CLI surface
```
emily-vanark, line 85: I think I expected this to be... more different?  Like, where do I tell it what I'm evaluating?

docs/architecture.md
```diff
@@ -0,0 +1,207 @@
...
+
+By default, generation writes under `output/` (or a user-specified parent):
+
+```text
+output/
+└── p_<user>__a_<agent>__t<turns>__r<runs>__<timestamp>/
+    ├── conversations/
+    │   ├── *.txt
+    │   └── logs/
+    └── evaluations/
```
emily-vanark: Is it possible we want a layer in between evaluations and specific judge runs that groups the judge runs by rubric / evaluator?  So `evaluations/SI/j_....` vs...

docs/architecture.md
```diff
@@ -0,0 +1,207 @@
...
+        └── j_<judge>__.../
+            ├── results.csv
+            ├── logs/
+            └── scores/          ← created by vera score
+```
+
+Pooled runs (`j_pooled__.../`) sit alongside `p_*` folders when merging multiple judge outputs. Full naming rules, legacy flat folders, and `output/adhoc` behavior: [README.md](../README.md).
```
emily-vanark: Is this pooling mutliple judges for the same conversations?  If so, I think I'd expect it to be at the level of conversations and evaluations within p_*.   If it's something else... what is it, please?

docs/vera-cli-use-cases.md
```diff
@@ -0,0 +1,70 @@
...
+updated: 2026-07-10
+---
+
+# `vera.py` — Use Cases for Review
+
+This document lists the intended use cases for the future `vera.py` entry point, so we can confirm nothing is missing before designing the CLI surface and config schema in detail.
+
+**Status: draft, open for review.** Please flag anything missing or wrong.
+
+## Use case 1 — End-to-end test of one LLM
```
emily-vanark, line 35: ... on a single evaluator?  On multiple evaluators?

docs/vera-cli-use-cases.md
```diff
@@ -0,0 +1,70 @@
...
+
+Generate -> judge -> score chained, for exactly one provider/agent LLM under test, in a single invocation.
+
+```
+vera pipeline --config run.json
+```
+
+**Open question:** should this ever support multiple providers natively (one combined comparison report), or is comparing providers always done by invoking this once per provider (external loop)?
```
emily-vanark: I... think it depends whether we think folks are going to have lots of hyperparamters for the different models, and whether we're sticking with command line arguments instead of a config file.  I think it would be nice to be able to easily specify, "run on models X, Y, and Z with these hyperparameters"... but I don't think that's user-friendly with command line args.

docs/architecture.md
```diff
@@ -0,0 +1,207 @@
...
+|------------|--------------|---------|
+| `vera generate` | `generate_conversations.runner` | Simulate conversations → `p_*` run folder |
+| `vera judge` | `judge.runner` | Evaluate transcripts → `j_*` evaluation folder |
+| `vera score` | `judge.score` | Aggregate `results.csv` → scores and visualizations |
+| `vera pool` | `judge.pool` (target) | Merge multiple judge runs |
+| `vera pipeline` | orchestration layer | Full workflow; passes paths between steps |
+
+Recommended invocations:
+
+```bash
```
emily-vanark, line 102: As I looka t the cli-use-cases below... I kind of long for a config file rather than the endles command line arguments we end up with in this model...   could that be part of the architecture?

docs/vera-cli-use-cases.md
```diff
@@ -0,0 +1,70 @@
...
+
+## Use case 2 — Batch generate across personas
```
emily-vanark, line 49: Are these... all personas in a single persona file?  Or potentially multiple persona files (and persona prompts) for mutliple evaluators?

docs/vera-cli-use-cases.md
```diff
@@ -0,0 +1,70 @@
...
+
+```
+vera generate --config run.json
+```
+
+Comparing across *providers* (not personas) is done by looping this invocation externally, once per provider — not a built-in cross-product flag.
+
+## Use case 3 — Judge existing conversations
+
+Judge one **or more** existing transcript folders against one or more rubrics. Each rubric has a default judge-LLM set, overridable per rubric. Multiple judges per rubric are supported (for judge-agreement analysis).
```
emily-vanark: Do you want to force judges to work only on conversations generated from their intended personas?  Or can those be decoupled?

docs/vera-cli-use-cases.md
emily-vanark: As I read more closely...  I do see there's a `--config run.json` here - should we delineate what goes in that at all in the architecture.md?

docs/vera-cli-use-cases.md
```diff
@@ -0,0 +1,70 @@
...
+
+```
+vera judge --conversations p_run_a/ p_run_b/ --config run.json
+```
+
+**Open question:** when judging 2+ folders together on the same rubrics, is the output one combined comparison result (side-by-side per-folder scores), or independent per-folder `results.csv` (batching as convenience only, no cross-folder aggregation)? This is a separate open question from use case 1's — resolving one does not resolve the other.
```
emily-vanark: I feel like keeping them separate is cleaner and then letting the user concat is kind of trivial.  But curious what @jgieringer thinks.

docs/vera-cli-use-cases.md
```diff
@@ -0,0 +1,70 @@
...
+
+## Open questions summary
+
+1. Multi-provider pipeline: native support in `vera pipeline`, or always an external loop? (Use case 1)
+2. Multi-folder judge output: combined comparison, or independent per-folder results? (Use case 3)
+
+## Not yet covered — flag if needed
+
+- Dry-run / validate-only (check config and inputs resolve without spending any LLM calls)
+- Resuming a partially-completed run
```
emily-vanark: I think figuring out how to gracefully resume a run is key - @jgieringer spent a lot of time trying to de-pain this process for the VERA v1 codebase.  We need it built in here from the start.

docs/vera-cli-use-cases.md
```diff
@@ -0,0 +1,70 @@
...
+- Listing available personas/rubrics/judges
```
emily-vanark: This seems helpful

docs/architecture.md
luca-belli: Sorry are you referring to this diagram? https://github.com/SpringCare/VERA-MH/pull/170/changes#diff-140eef3ba41bdcf401d507408084181f2c0ac627532b61e0f7906ea7cc926782R19

For now, no major changes in the strture, but we need to reconsider naming, as they are getting a little cumbersome.

docs/architecture.md
luca-belli: good call out. those are not going to be inside `src` as I want them to be accessible to everyone. All the files in src can of course reference those

docs/architecture.md
luca-belli: I am not sure if I understand the question. I only meant that there is a top level script, which will dispatch work accordingly (see the cli use cases file)

docs/architecture.md
luca-belli: yes. my current thinking is to support both (specifically, we support CLI and if you have a config file it gets expanded on the CLI automatically)

the reason is to avoid sending files back and forth when/if we make it as a service
I am also a fan of config files, but CLI are cleaner because they are immutable so they make it easier to understand what exactly was run when (e.g. `vera config.json` gives me less info, as config might have changed)

docs/architecture.md
luca-belli: oh good point. Will think this through. 
and / or the j_ part contains that kind of info as well

docs/architecture.md
luca-belli: I think this is a hack that @jgieringer and I discussed some time ago to make it fit use cases we didnt support. It pools multiple generated convos and uses the same judge

docs/vera-cli-use-cases.md
luca-belli: both/either 
will specify thanks

docs/vera-cli-use-cases.md
luca-belli: absolutely. this is the next step after the top level API is finalized. I'll a note to say that the config and the CLI are going to respect the same set of parameters

docs/vera-cli-use-cases.md
luca-belli: Good question, and I will specify. I think multiple files, with the case of single file as well. Each file contains multiple personas. There might be some duplicate personas

docs/vera-cli-use-cases.md
luca-belli: Those can be decoupled. The human user will point to the set of conversations that are required

docs/vera-cli-use-cases.md
luca-belli: There are also in-between options eg separate, but the scoring layers aggregates them

docs/vera-cli-use-cases.md
luca-belli: Will add thank you

docs/vera-cli-use-cases.md
jgieringer: I think outputting some sort of run-config.json will be crucial to make this less painless. AND should be a very trivial thing to add.
It could even be something like:
`vera resume --config path/to/output/run-config.json`

the run-config.json would store the initial run's config plus # of errors, paths things were stored, your api_key (jk), etc.

and given the metadata stored, it "knows" exactly what needs resuming.

docs/vera-cli-use-cases.md
jgieringer: reminds me that we want to support multiple user models
`... -u gpt-5.4:2 claude-sonnet-5:1 ...`

this also reminds me of a bug: with LLMs having different params (reasoning_effort vs thinking_effort), -jep gets applied to all of the -j models
so `-j gpt-5.4:2 claude-sonnet-5:3 -jep reasoning_effort=low` would try setting reasoning_effort to sonnet and might get an error from claude's api due to unknown param.

Creating an issue to follow-up on this.

jgieringer: https://github.com/SpringCare/VERA-MH/issues/172

docs/architecture.md
jgieringer: great point! Luca and I talked about this a while ago and here are our thoughts
https://github.com/SpringCare/VERA-MH/issues/173

docs/architecture.md
jgieringer: #173

docs/architecture.md
jgieringer: all the more reason to _also_ output a run-config.json per vera call? That way if something errors, you can enter `vera resume path/to/run-config.json` and the run-config.json contains alllll the info?

docs/architecture.md
jgieringer: yeah this is interesting!
I wonder if we want an overall parent+timestamp folder
small plug for [coolname](https://pypi.org/project/coolname/) 😂 

something like
```
output/
    └── prophetic-bullfrog-20260713/
    |     └──── run-config.json
    │     └──── SI/
    │     └────── conversations/
    │     │     ├── *.txt
    │     │     └── logs/
    │     └────── evaluations/
    │     │    └────── j_.../
    │     │    │     ├── *.tsv
    │     │    │     ├── results.csv
    │     │    │     ├── scores/
    │     │    │     └── logs/
    │     │    └────── j_.../
    │     │    │     └── ...
    │     │    └────── j_pooled/
    │     │    │     ├── results.csv
    │     │    │     └── scores/
    └──── HFO/
          └────── ...
```

but having some of the config in the name is helpful.. so
```
output/
    └── p_<user>__a_<agent>__t<turns>__r<runs>__<timestamp>/
    |     └──── run-config.json
    │     └──── SI/
    │     └────── conversations/
    │     │     ├── *.txt
    │     │     └── logs/
    │     └────── evaluations/
    │     │    └────── j_.../
    │     │    │     ├── *.tsv
    │     │    │     ├── results.csv
    │     │    │     ├── scores/
    │     │    │     └── logs/
    │     │    └────── j_.../
    │     │    │     └── ...
    │     │    └────── j_pooled/
    │     │    │     ├── results.csv
    │     │    │     └── scores/
    └──── HFO/
          └────── ...
```

docs/architecture.md
jgieringer: yeah this was something I overlooked.. the pooled judgings by default go to `/output` instead of the `p_...` dir 🤦 
I would like it to be something like what I put in another comment
```
output/
    └── p_<user>__a_<agent>__t<turns>__r<runs>__<timestamp>/
    |     └──── run-config.json
    │     └──── SI/
    │     └────── conversations/
    │     │     ├── *.txt
    │     │     └── logs/
    │     └────── evaluations/
    │     │    └────── j_.../
    │     │    │     ├── *.tsv
    │     │    │     ├── results.csv
    │     │    │     ├── scores/
    │     │    │     └── logs/
    │     │    └────── j_.../
    │     │    │     └── ...
    │     │    └────── j_pooled/
    │     │    │     ├── results.csv
    │     │    │     └── scores/
    └──── HFO/
          └────── ...
```

jgieringer (commented):
First pass through. This looks great.
I want to gander again tomorrow.
I love having `vera this` `vera that` ❤️ 

This wouldn't be merged now, though, right??

docs/architecture.md
luca-belli: yes absolutely. this is the direction it's going

docs/architecture.md
luca-belli: tbh i think a lot of that will be driven by testing 
and i am trying to get rid of pooled judges but maybe that's impossible

docs/vera-cli-use-cases.md
luca-belli: thank you, that's very valuable info and will influence the design 
i think we should keep the extra parameters as close as possible to their original names to avoid confusion

.github/CODEOWNERS
```diff
@@ -0,0 +1,17 @@
+# Architecture and domain packages require maintainer review.
+/docs/architecture.md @SpringCare/vera-mh-maintainers
```
emily-vanark, line 2: Does this maintainers group exist?  I couldn't find it, but my github skillz may have been lacking.

emily-vanark:
Remind me why we're merging to main instead of a VERA_2.0 branch?

docs/vera-cli-use-cases.md
```diff
@@ -0,0 +1,156 @@
...
+
+Generate -> judge -> score chained, for exactly one chatbot under test, in a single invocation.
+
+```
+vera pipeline --config run.json
+```
+
+**Resolved:** stays single-chatbot-per-invocation. Comparing chatbots is always an external loop over single-chatbot pipeline runs, consistent with use case 2. Native multi-chatbot support (one combined comparison report) is not built now — flagged as a possible future addition if a real need emerges, not ruled out permanently.
+
+**`--target <name>` shorthand:** `vera pipeline --target SI` resolves `SI` to one rubric bundle manifest (see [Rubric bundle manifest](../architecture.md#rubric-bundle-manifest)) and sets *both* the generation personas and the judging rubric from it in one shot — for the common case of "run the canonical test for X." This is the one deliberate exception to personas/rubrics being chosen independently; every other invocation (`--rubric` plus separately-specified personas, or explicit `--config` blocks) keeps generation and judging fully orthogonal.
```
emily-vanark: So, does this mean that if we don't specifify --target <name> we run... all the evaluators?  Just one default evaluator?  (which?)

docs/vera-cli-use-cases.md
emily-vanark: I'm still not clear, reading this, on the answer to my first question in this thread.  I don't see it explicitly saying anywhere, "this will run all the evaluators" or "this will run only  one evaluator" or "this will run all the evaluators specified in the run.json config file" or... ?   I'm still confused.

docs/vera-cli-use-cases.md
emily-vanark: This is another one where I don't actually see the answer specified below...   I still deeply do not understand what persona file(s) would be run by `vera generate gpt:1 sonnet:2`...

emily-vanark: Or, for the matter, what LLM(s?) would be evaluated by `vera generate gpt:1 sonnet:2`....  is there a default config that's specifying everything not specified on the command line, even if we don't use `--config`?  Are there a bunch of defaults hidden in the vera.py parser or the files it calls?

emily-vanark: Do we need both lines  35 and 36 to make this work?  They... don't look like a single command to me, right now.  But maybe that's how they're meant?

docs/vera-cli-use-cases.md
```diff
@@ -0,0 +1,156 @@
...
+
+Personas come from one or more persona files, each containing multiple personas; duplicate persona names across files are possible (disambiguated by file + name, see Naming).
+
+## Use case 3 — Judge existing conversations
+
+Judge one **or more** existing transcript folders against one or more rubrics. Each rubric has a default judge-LLM set, overridable per rubric. Multiple judges per rubric are supported (for judge-agreement analysis). Judging is decoupled from generation — point it at whatever conversation folders you need, with no enforced coupling to originating personas.
+
+```
+vera judge --conversations output/c_sonnet/<run>/conversations/ --config run.json
+vera judge -j claude:1 gpt:2
```
emily-vanark: If line 51 is a standalone call, I don't understand how we know what conversations we're judging, with which rubric.

docs/vera-cli-use-cases.md
```diff
@@ -0,0 +1,156 @@
...
+```
+vera pool --evaluations <folder> <folder> ...
+```
+
+Owned by `score/`, consistent with score owning aggregation-across-runs (as opposed to judging's own within-run aggregation). This formalizes and generalizes the `vera pool` subcommand as a first-class, general-purpose capability.
+
+## Use case 6 — Resume
+
+```
+vera resume --config output/c_sonnet/<run>/config.json
```
emily-vanark, line 121: Is this equally valid whether what needs to resume is generate, judge, or score?

docs/vera-cli-use-cases.md
```diff
@@ -0,0 +1,156 @@
...
+- **CLI flags and `--config` are strictly either/or, never combined for the same run.** A given piece of information (model selection/repeats, sampling knobs, persona/rubric lists) is supplied via one or the other, never both — the implementation rejects the combination rather than silently merging.
+- Internally, `--config` always resolves to the same canonical flag-set the CLI would produce, so there is exactly one resolved form regardless of input path. The tool prints this resolved form at run start for terminal/CI-log visibility (it does not write to the shell's own history — an opt-in `--print` flag emits the resolved flag-string with no execution, for a caller who wants to `eval` it into their own shell explicitly).
+- JSON, not YAML — robust when passed as a one-line env var or stdin payload with no escaping ambiguity.
+
+### `config.json` shape
+
+Top-level `generation` and `judging` blocks are **completely orthogonal** — model selection for one must never influence or be influenced by the other. Each follows the same models-list pattern (a list, not an object keyed by name, so the same model can appear twice with different knobs):
```
emily-vanark: Where is the chatbot to be evaluated specified?  In the config.json?  If so... where / how?  Somewhere else?

docs/architecture.md
```diff
@@ -0,0 +1,288 @@
...
+                                   judge ──► c_<chatbot>/<run>/evaluations/<rubric>/j_*/results.csv
+                                        │
+                                        ▼
+                                   score ──► .../j_*/scores/
+                                        │
+                                        ▼
+                                   pool ──► pooled result across evaluation folders
+```
+
+## Domain model
```
emily-vanark, line 37: Does the chatbot being evaluated belong in the Domain model here?

emily-vanark: (I feel like maybe if it were here, it would have percolated through to being included in the config.json example?)

docs/architecture.md
```diff
@@ -0,0 +1,288 @@
...
+| `vera score` | `score.score` | Aggregate `results.csv` → scores and visualizations |
+| `vera pool` | `score.pool` | Concatenate multiple evaluation folders into one pooled result |
+| `vera pipeline` | orchestration layer | Full workflow for one chatbot; passes paths between steps |
+| `vera resume` | orchestration layer | Reads `config.json` (sha-verified) + `state.json`, continues an incomplete run |
+
+Config and CLI flags are strictly either/or, never combined for the same run — see [vera-cli-use-cases.md](./vera-cli-use-cases.md#config-mechanism). `-u`/`-j` shorthand selects models/repeats for the user/judge side respectively; bespoke sampling knobs are config-only. `--rubric` selects the rubric-bundle manifest (see [Rubric bundle manifest](#rubric-bundle-manifest) below) — always a list, though only a length-1 list is supported until Phase 4.
+
+```bash
+uv run python vera.py pipeline --config run.json
+uv run python vera.py generate -u gpt:1 sonnet:2
```
emily-vanark: what's the -c model here?  or does it not need to be specified?

docs/architecture.md
```diff
@@ -0,0 +1,288 @@
...
+```json
+{
+  "rubric_file": "rubric.tsv",
+  "rubric_prompt_beginning_file": "rubric_prompt_beginning.txt",
+  "question_prompt_file": "question_prompt.txt",
+  "personas": ["data/personas.tsv"]
+}
+```
+
+Paths are relative to the manifest's own folder. `personas` is **informational only** — it documents which personas this rubric is intended/validated for, for humans and tooling to discover; it does not make generation consume it automatically. Generation still chooses personas independently (the `generation`/`judging` orthogonality invariant holds). This manifest shape is exactly what a `judging.rubrics[]` config entry looks like once Phase 3 formalizes the schema — the format isn't thrown away when the CLI is replaced, it's the design.
```
emily-vanark: So does the personas path have a different "relative to" than the rest of the items in the list?  That's... kind of confusing... wonder if there's any way to make it obvious within the manifest?  (Does json support comments?)

docs/architecture.md
```diff
@@ -0,0 +1,288 @@
...
+- Overwrite an existing run's output folder silently — collision on an already-existing folder errors out (no overwrite, no auto-suffix).
+- Bypass architecture checks (pyright, required CI) to merge structural changes.
+
+### Stable interfaces (agent-coding optimization)
+
+This codebase is optimized for agent coding: most files are safe for an agent to change freely within a package's own boundaries, but a small set of **stable interfaces** are rarely meant to change and require a design doc before modification, not just a PR. These are called out individually in [`.github/CODEOWNERS`](../.github/CODEOWNERS) (not just covered by their package's blanket rule) so their significance is visible at a glance:
+
+| File | What it stabilizes |
+|------|---------------------|
+| `llm_clients/llm_interface.py` | `LLMInterface` ABC — every provider implements this |
```
emily-vanark: I don't know what ABC means here.

.github/CODEOWNERS
luca-belli: it does now!

luca-belli:
> Remind me why we're merging to main instead of a VERA_2.0 branch?

This is to show/start enforcing the new rules. Development phases will be added to a new branch

docs/architecture.md
luca-belli: added thank you

docs/vera-cli-use-cases.md
luca-belli: to answer your question: `--target` (or equivalent in the config file) 

There is no default option, to prevent launching too big jobs

I do suspect that at some point we'll define shorthands for commonly used groups of evals (e.g., `clinical` or whatnot)

docs/vera-cli-use-cases.md
luca-belli: rephrased below, and it should answer your question now

docs/vera-cli-use-cases.md
luca-belli: yes. but i suspect score won't require as much resuming

docs/vera-cli-use-cases.md
luca-belli: Should be clarified below now

docs/ARCHITECTURE-SPINE.md
```diff
@@ -0,0 +1,362 @@
...
+
+This is not hexagonal/ports-and-adapters in the strict sense: domain packages import concrete infrastructure packages directly (`generate/` and `judge/` call into `llm_clients/` and `storage/` themselves), rather than depending only on domain-owned ports that outer adapters implement via injection. What it borrows from hexagonal is narrower and more useful here: each infrastructure concern owns its own interface next to its own implementations (`llm_clients/` holds `LLMInterface` plus every provider, `workers/` holds `QueueProtocol` plus `LocalQueue`/`SQSQueue`, `storage/` holds `StorageBackend` plus `LocalFilesystemStorage`/`S3Storage`), instead of a single cross-cutting `interfaces/` package grouping every ABC by "abstract vs. concrete." That colocation is chosen explicitly (memlog) to satisfy the Common Closure Principle: an LLM-provider change and a storage-backend change have nothing to do with each other and should not force a shared package to move.
+
+Layer → package mapping:
+
+| Layer | Packages |
+| --- | --- |
+| CLI orchestrator | `vera.py` |
+| Domain (mutually isolated) | `generate/`, `judge/`, `score/` |
```
jgieringer, line 33: probably under `judge/` but if reporting evolved to utilize LLMs to provide creative suggestions for chatbot improvement, there could be a `report/` package

docs/ARCHITECTURE-SPINE.md
```diff
@@ -0,0 +1,362 @@
...
+
+### AD-23 — Output layout is nested-only, with path-first stage contracts
+
+- **Binds:** `generate/`, `judge/`, `score/`, `utils/naming.py`
+- **Prevents:** stage code depending on a flat-layout path shape that no longer exists; resume logic re-deriving state from scratch instead of reading it.
+- **Rule:** [ADOPTED] The flat output layout is removed for *new writes*; only the nested layout (`output/c_<chatbot>/<run>/conversations/`, `.../evaluations/<rubric>/j_<run>/...`) is valid for a run created going forward. This does not remove the ability to *read* existing flat-layout data: `vera judge --conversations <folder>` and `vera score -r <results.csv>` take an explicit path and read files by their own format, never by re-deriving meaning from the parent folder's naming pattern, so both keep working against old-layout output. `vera resume` is the one path that genuinely can't operate on an old run, since `config.json`/`.sha256`/`state.json` didn't exist under the old layout — there is nothing for it to read, independent of naming scheme. Stage handoffs are path-first contracts, recording models/personas/artifact paths/timestamps, usable as input for resume or a new run.
```
jgieringer, line 223: I wonder in the output of both conversations/... and evaluations/rubric/j_... if we should have an errors.json which stores the persona, error, etc., that way we can store errors, and on `vera resume` it retries those specific fails instead of having to do math by parsing txt/tsv files.
On resume, the errors are either cleared from the json file or updated with the latest error(s).

docs/ARCHITECTURE-SPINE.md
```diff
@@ -0,0 +1,362 @@
...
+        <rubric_name>/
+          j_<judge>_<nickname>_<timestamp>_<sha>/  # judge stays flat, no persistent parent
+            config.json
+            config.json.sha256
+            state.json
+            results.csv
+            scores/                                # created by `vera score`
+  evaluations/
+    <config-sha256>/                                # persistent per-config directory, same role as c_<chatbot>/
+      <rubric_name>/                                # standalone judging (vera judge run on its own)
```
jgieringer, line 304: I wonder if conversations/u_<persona-file>... would get really large?
conversations/\<persona-file\>/... ?
then have
evaluations/<rubric_name-persona-file>/...
in case you wanted to run same rubric on different personas?

jgieringer (commented):
Just added some thoughts, nothing pressing!

docs/ARCHITECTURE-SPINE.md
luca-belli: oh interesting, that's a good idea. I think we can add that by adding saying what kind of questions were failed on

docs/ARCHITECTURE-SPINE.md
luca-belli: tbh I am not sure myself. I think there are good argument to be made for each case, but only using it a bunch we'll figure out where the problem lies and how do we might want to change it

Would you be ok with this setup, with the understanding that can (and should) evolve over time?

docs/ARCHITECTURE-SPINE.md
luca-belli: The way I was thinking is that if a file is not in the json manifest of completed files it should be treated as an error or a non complete case, with logs logging the error

jgieringer (approved):
Very thoughtful and thorough work! LGTM 🙇

docs/ARCHITECTURE-SPINE.md
emily-vanark: So... that report that Josh currently has code to generate... that could put its output in `report/`?

emily-vanark (approved):

docs/ARCHITECTURE-SPINE.md
luca-belli: yes, I added a note about it: https://github.com/SpringCare/VERA-MH/pull/170/changes/f82be4c63b1087390441cebe22bf2204992515f1#diff-140eef3ba41bdcf401d507408084181f2c0ac627532b61e0f7906ea7cc926782R155

PR #171, opened 2026-07-09
Support reasoning/thinking levels in LLMs

jgieringer:
## Description
This PR adds support & testing for reasoning across major LLMs


llm_clients/claude_llm.py
```diff
@@ -24,6 +24,20 @@
...
 _DEFAULT_ANTHROPIC_CACHE_CONTROL: Dict[str, Any] = {"type": "ephemeral"}
 
+# Models that use `thinking={"type": "adaptive"}` + the `effort` field instead
+# of `thinking={"type": "enabled", "budget_tokens": N}`. Also the set of
+# models that reject `temperature` outright (see `_unsupported_model_params`).
+_ADAPTIVE_THINKING_MODEL_MARKERS = ("opus-4-8", "sonnet-5")
+
+# thinking_effort labels -> Anthropic's raw `budget_tokens`, for models that
+# don't support the `effort` shorthand (haiku, older sonnet).
+_EFFORT_BUDGETS: Dict[str, int] = {
```
luca-belli, line 66: could you document where those numbers come from? and whether we need to update them over time?

llm_clients/claude_llm.py
```diff
@@ -24,6 +24,56 @@
 # and be removed from memory if it is not used.
 _DEFAULT_ANTHROPIC_CACHE_CONTROL: Dict[str, Any] = {"type": "ephemeral"}
 
+# Per-model quirks, keyed by a substring marker matched against model_name.
```
luca-belli, line 27: is there a documentation page we could add?

llm_clients/claude_llm.py
```diff
@@ -34,9 +84,45 @@ class ClaudeLLM(JudgeLLM):
 
     def _unsupported_model_params(self) -> Dict[str, frozenset[str]]:
         return {
-            "opus-4-8": frozenset({"temperature"}),
+            marker: frozenset({"temperature"})
```
luca-belli: is this discarding temperature?

llm_clients/llm_interface.py
```diff
@@ -65,7 +65,21 @@ def ensure_pydantic_response(response: Any, response_model: Type[T]) -> Optional
...
         except ValidationError:
-            return None
+            pass
+        # Some models (observed on claude-sonnet-5) occasionally wrap forced
+        # tool-call args in a single spurious key instead of the real field
+        # names - e.g. {"$PARAMETER_NAME": {"answer": ..., "reasoning": ...}}
+        # or {"$PARAMETER_VALUE": {...}}, echoing Anthropic's legacy XML
+        # tool-use template placeholders. If there's exactly one key and its
+        # value is itself a dict, retry against that inner dict.
+        if len(response) == 1:
```
luca-belli: shouldn't this be inside the ValidationError block?

README.md
```diff
@@ -361,6 +361,30 @@ uv run python judge.py -f output/{YOUR_P_RUN} -j gpt-4o:3
...
+```bash
+# Judge with high reasoning effort
+uv run python judge.py -f output/{YOUR_P_RUN} -j gpt-5.4 -jep reasoning_effort=high
+
+# Claude judge with extended thinking
+uv run python judge.py -f output/{YOUR_P_RUN} -j claude-sonnet-5 -jep thinking_effort=high
+```
+
+**Notes:**
+- `thinking_effort` is Claude-specific shorthand (not a native Anthropic API field): it's translated internally into the right combination of `thinking`/`effort`/`budget_tokens` for the given model (see [`llm_clients/claude_llm.py`](llm_clients/claude_llm.py)). It exists because the CLI's extra-params parser splits on commas, so a literal multi-key dict like `thinking={"type": "enabled", "budget_tokens": 5000}` can't be passed directly.
```
luca-belli, line 384: to clarify: this is a limitation of OUR cli, correct?

jgieringer: correct, wasn't a simple `-jep foo=bar` to enable thinking, so added this to support thinking across claude models

llm_clients/claude_llm.py
jgieringer: it gets filtered out, not removed, at line 198 then passed to the langchain constructor!

llm_clients/llm_interface.py
jgieringer: ohhh nice catch! fixed

llm_clients/llm_interface.py
```diff
@@ -65,6 +65,19 @@ def ensure_pydantic_response(response: Any, response_model: Type[T]) -> Optional
...
             return response_model.model_validate(response)
         except ValidationError:
+            # Some models (observed on claude-sonnet-5) occasionally wrap forced
+            # tool-call args in a single spurious key instead of the real field
+            # names - e.g. {"$PARAMETER_NAME": {"answer": ..., "reasoning": ...}}
+            # or {"$PARAMETER_VALUE": {...}}, echoing Anthropic's legacy XML
+            # tool-use template placeholders. If there's exactly one key and its
+            # value is itself a dict, retry against that inner dict.
+            if len(response) == 1:
+                key, inner = next(iter(response.items()))
```
luca-belli, line 75: nit: i don't think iter is needed here?

README.md
luca-belli: ok good to know, because we can always change things on our side :)

luca-belli (approved):

llm_clients/llm_interface.py
jgieringer: next(iter(...)) grabs that first item, so keeping as-is since this is a fall-back to try and pull wonky response data

PR #174, opened 2026-07-16
Release/v1.2

jgieringer:
## Summary

VERA-MH v1.2.0 — clinician-informed rubric revision, a new recommended judge profile, and reasoning/extended-thinking support across all three LLM providers.

- **Rubric v1.2** ([`data/rubric.tsv`](data/rubric.tsv)) — sharpened risk-language boundaries, clarified context for crisis-resource/distress-tolerance expectations, and reduced double-penalization across dimensions. 85% raw agreement (IRR = 0.79) with human clinician raters on a 40-conversation validation set.
- **Recommended judge** — GPT 5.4 (`gpt-5.4`) with `reasoning_effort=low`, now the default in `scripts/run_recommended_vera_pipeline.sh`.
- **Reasoning support** — `llm_clients/` now handles OpenAI `reasoning_effort`, Claude `thinking_effort`, and Gemini `thinking_level`/`thinking_budget`, including per-model quirk handling (e.g. Claude adaptive-thinking models, Gemini 3+/GPT-5+ sampling-param rejection) and structured-output resilience for Claude's occasional wrapped tool-call responses.
- **Docs** — `docs/evaluating.md` gains a reasoning-by-provider table and a Claude per-model quirks table; `AGENTS.md`/`CLAUDE.md` consolidated; `LICENSE` clarified for research/benchmarking use only.

Full details in [CHANGELOG.md](CHANGELOG.md#v120).

## Breaking / migration

- Scores from the v1.2 rubric are not comparable to prior versions — re-judge existing conversations against `data/rubric.tsv` before comparing.
- Automation pinned to a different judge should be updated to GPT 5.4 / `reasoning_effort=low`.

## Test plan

- [x] `uv run pytest -m "not live"` passes
- [x] Spot-check a re-judged conversation set against the v1.2 rubric for sane scores
- [x] Confirm `scripts/run_recommended_vera_pipeline.sh` runs end-to-end with the new default judge config


judge/score_utils.py
```diff
@@ -125,7 +125,7 @@ def pct_of_total(count: int | float, total: int | float, decimals: int = 2) -> f
 
 def calculate_vera_score(bp_pct: float, hph_pct: float) -> float:
     """
-    Calculate VERA-MH v1.1 score using the standard formula.
+    Calculate VERA-MH v1 score using the standard formula.
```
emily-vanark: Do we want this to be v1 or v1.2?

emily-vanark (approved):
Picked one nit, but all the tests passed and overall looks great!

Um... is there a release page with the new General LLM scores, or does that go somewhere else?

judge/score_utils.py
jgieringer: I'm wondering about even just removing the vX.X score all-together so we don't have to keep updating it 😅 
In the meantime, will put v1.2, but the scoring formula hasn't changed since v1. BUT to keep with the graphics, v1.2 is most consistent methinks.

PR #177, opened 2026-07-22
feat: add RubricConfig.load_bundle() for rubric bundle manifests

luca-belli:
## Summary

**Phase 0 foundation, for both generation and judging** (docs/architecture.md's Phase 0 migration entry, expanded in PR #170 to cover both sides of the pipeline, not just judging). This PR only builds the shared, reusable pieces -- no CLI flag is wired up here, on either side. PR #178 does the wiring for both.

- Adds `utils/rubric_manifest.py` (`load_manifest()` / `load_manifest_personas()`), a manifest JSON reader in `utils/` (leaf layer) so **either** `judge/rubric_config.py` **or** `generate.py` can read a rubric bundle manifest without importing from each other's package (`generate/` ⊥ `judge/`, per docs/architecture.md's Layer model). `load_manifest()` returns the rubric-file fields judging needs; `load_manifest_personas()` returns the `personas` list generation needs -- same manifest, same reader, two accessors.
- Adds `RubricConfig.load_bundle()`, the judging-side consumer of that reader: a classmethod that loads a rubric from a rubric bundle manifest (delegates to the existing `RubricConfig.load()`). This is judging's half of the foundation -- generation's half is just `load_manifest_personas()` itself, since `generate.py` doesn't need a whole `RubricConfig`-equivalent object, just the persona list.
- Adds `data/rubric_manifest.json`, the production default manifest pointing at the existing `data/rubric.tsv` bundle (and, once #178 wires it up, the personas generation will use for that rubric)
- Adds `tests/fixtures/rubric_manifest_simple.json`, a test fixture manifest for the existing `rubric_simple.tsv` fixture
- **Fix (added after initial review):** `load_manifest_personas()` was returning a manifest's `personas` entries verbatim instead of resolving them relative to the manifest's own folder, contradicting docs/architecture.md's stated rule for manifest paths (the same rule `rubric_file`/etc. already follow via `RubricConfig.load()`). Caught while reorganizing `data/` on a downstream branch and fixed here instead, since this PR hadn't merged yet and is where the bug originated. `data/rubric_manifest.json`'s `personas` entry updated to the correct manifest-relative form (`"personas.tsv"`, not `"data/personas.tsv"`) now that resolution actually happens; both relative and absolute-path entries are covered by tests.

**Not in this PR:** `judge.py --rubrics` still hardcodes `RubricConfig.load(rubric_folder="data")`, and `generate.py` still hardcodes `data/personas.tsv` -- neither is wired to the manifest yet. That's PR #178.

## Test plan
- [x] `uv run pytest -m "not live"` passes (970 passed)
- [x] Unit tests cover: successful load via manifest, `FileNotFoundError` on missing manifest, `ValueError` on missing required manifest key -- both on `RubricConfig.load_bundle()` and directly on the shared `utils.rubric_manifest` module (covering both `load_manifest()` and `load_manifest_personas()`, including relative and absolute persona-path entries)
- [x] `uv run ruff check` / `uv run ruff format --check` clean
- [x] `uv run pyright` clean (no new errors vs. base)

### DIFF
--- a/data/rubric_manifest.json
+++ b/data/rubric_manifest.json
@@ -2,5 +2,6 @@
   "rubric_file": "rubric.tsv",
   "rubric_prompt_beginning_file": "rubric_prompt_beginning.txt",
   "question_prompt_file": "question_prompt.txt",
-  "personas": ["personas.tsv"]
+  "personas": ["personas.tsv"],
+  "persona_context_template_file": "persona_context_template.txt"
 }
--- a/data/rubric_manifest.json
+++ b/data/rubric_manifest.json
@@ -1,0 +1,6 @@
+{
+  "rubric_file": "rubric.tsv",
+  "rubric_prompt_beginning_file": "rubric_prompt_beginning.txt",
+  "question_prompt_file": "question_prompt.txt",
+  "personas": ["data/personas.tsv"]
+}
--- a/data/rubric_manifest.json
+++ b/data/rubric_manifest.json
@@ -2,5 +2,5 @@
   "rubric_file": "rubric.tsv",
   "rubric_prompt_beginning_file": "rubric_prompt_beginning.txt",
   "question_prompt_file": "question_prompt.txt",
-  "personas": ["data/personas.tsv"]
+  "personas": ["personas.tsv"]
 }
--- a/utils/rubric_manifest.py
+++ b/utils/rubric_manifest.py
@@ -1,0 +1,67 @@
+"""Shared rubric bundle manifest reading.
+
+A rubric bundle manifest (docs/architecture.md#rubric-bundle-manifest)
+attaches a rubric and the personas it's validated for as one unit. Both
+`generate.py` (personas half) and `judge/rubric_config.py` (rubric half)
+read the same manifest file -- this lives in `utils/` (the leaf layer)
+rather than in `judge/` so `generate.py` never has to import a `judge/`
+module to read it (`generate/`/`judge/` must never import each other, per
+docs/architecture.md's Layer model).
+"""
+
+from __future__ import annotations
+
+import json
+from pathlib import Path
+from typing import Any
+
+import aiofiles
+
+REQUIRED_KEYS = (
+    "rubric_file",
+    "rubric_prompt_beginning_file",
+    "question_prompt_file",
+)
+
+
+async def load_manifest(manifest_path: str) -> dict[str, Any]:
+    """Read and validate a rubric bundle manifest JSON file.
+
+    Args:
+        manifest_path: Path to the rubric bundle manifest JSON file.
+
+    Returns:
+        The parsed manifest dict.
+
+    Raises:
+        FileNotFoundError: If the manifest doesn't exist.
+        ValueError: If the manifest is missing a required rubric key.
+    """
+    manifest_file = Path(manifest_path)
+    if not manifest_file.exists():
+        raise FileNotFoundError(f"Rubric bundle manifest not found: {manifest_file}")
+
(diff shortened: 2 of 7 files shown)


jgieringer (approved):
LGTM

data/rubric_manifest.json
```diff
@@ -0,0 +1,7 @@
+{
```
emily-vanark, line 1: Is/are the persona prompt(s) part of the bundle?

utils/rubric_manifest.py
```diff
@@ -0,0 +1,85 @@
...
+
+
+async def load_manifest_personas(manifest_path: str) -> list[str]:
+    """Read a rubric bundle manifest's `personas` list.
+
+    Used by `generate.py --rubric-manifest` (Phase 0's generation-side
+    counterpart to `judge.py --rubrics`, see docs/architecture.md's Phase 0
+    migration entry) to select personas from the same manifest that
+    `judge.py` loads the rubric from. `personas` is optional in the
+    manifest and defaults to an empty list.
```
emily-vanark, line 71: So if I run generate on a rubric manifest that doesn't define the personas, what happens?

utils/rubric_manifest.py
```diff
@@ -0,0 +1,85 @@
...
+    """
+    manifest = await load_manifest(manifest_path)
+    manifest_dir = Path(manifest_path).parent
+    return [str(manifest_dir / p) for p in manifest.get("personas", [])]
+
+
+async def load_manifest_persona_context_template(manifest_path: str) -> str:
+    """Resolve a manifest's persona context template relative to the manifest."""
+    manifest = await load_manifest(manifest_path)
+    context_template = manifest.get("persona_context_template_file")
```
emily-vanark, line 85: How is the context template used?

emily-vanark (approved):
Left a few qusetions, but LGTM

data/rubric_manifest.json
luca-belli: yes, it points to the existing persona file (see line 5 below)

luca-belli: This will help us define a `--target` CLI later

utils/rubric_manifest.py
luca-belli: that won't work
admittedly this is a little bit of a workaround to allow for rubric and persona choosing without touching the CLI (I am working on that in parallel)
However, I will make sure that this is taken into account in that work, and the right checksa re in place

utils/rubric_manifest.py
luca-belli: it's the part that reads the csv of the personas and makes a prompt out of it https://github.com/SpringCare/VERA-MH/pull/177/changes/BASE..72907825a2e9e63303361f8b61d6e14a70595468#diff-8c17a4e6f4b2665e056d05bc12902e31883bf4edbc2f71f27893005e6f638c57

Because it's specific to the columns of the personas, it should be in the metadata

utils/rubric_manifest.py
```diff
@@ -0,0 +1,85 @@
...
+        manifest = json.loads(await f.read())
+
+    missing_keys = [key for key in REQUIRED_KEYS if key not in manifest]
+    if missing_keys:
+        raise ValueError(
+            f"Rubric bundle manifest {manifest_file} is missing required "
+            f"key(s): {missing_keys}"
+        )
+
+    return manifest
```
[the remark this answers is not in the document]
luca-belli: good bot

data/rubric_manifest.json
luca-belli: although as you pointed out, it's not really robtust between generate and judge. Will improve for the CLI version

PR #180, opened 2026-07-29
refactor: move the SI rubric bundle into data/SI/

luca-belli:
## Summary

Stacked on #178 (base branch is that PR's branch, not `main`).

Everything currently in `data/` (`rubric.tsv`, `rubric_prompt_beginning.txt`, `question_prompt.txt`, `rubric_manifest.json`, `personas.tsv`) is specific to the SI rubric. Once a second rubric type (e.g. PHQ9) exists, `data/` needs a subfolder per rubric rather than one flat, implicitly-SI set of files.

- Moves all five files into `data/SI/`
- Updates every hardcoded default/reference across `generate.py`, `judge.py`, `run_pipeline.py`, `judge/score.py`, `judge/score_utils.py`, `scripts/pool_vera_scores.py`, `scripts/summarize_results.py`, `README.md`, `docs/judge.md`, and tests
- No behavior change: `data/SI/rubric_manifest.json`'s `personas` entry (`"personas.tsv"`) still resolves correctly since it's manifest-relative (per #177's fix) and the manifest and `personas.tsv` remain siblings after the move

## Test plan
- [x] `uv run pytest -m "not live"` passes (976 passed)
- [x] `uv run ruff check` / `uv run ruff format --check` clean (same pre-existing, unrelated drift as base -- nothing new)
- [x] `uv run pyright` clean (same pre-existing, unrelated error count as base -- nothing new)


jgieringer (approved):
LGTM

emily-vanark (approved):
LGTM

emily-vanark:
I think github copilot had a few good points we should address before merging...

luca-belli:
> I think github copilot had a few good points we should address before merging...

Thank you, should be addressed now

PR #184, opened 2026-07-31
feat(cli): add unified vera.py CLI orchestrator wired to domain services

luca-belli:
## Summary

This stacked PR adds the Phase 1 unified `vera.py` command surface on top of #180. It wires `generate`, `judge`, `score`, `pool`, and `pipeline` to parser-independent application functions; `resume` is reserved and fails explicitly until its state/checksum contract is implemented.

## Architecture decisions

- `feat/VERA_2.0` is the source of truth for the CLI architecture.
- A run is configured by either JSON config or command-line flags. `--sample` is the sole CLI flag permitted alongside `--config`.
- `vera.py` calls domain functions directly. It does not import or dynamically invoke the legacy `generate.py` or `judge.py` CLIs.
- Generation consumes resolved persona files, not rubric manifests. `--target` resolves persona files from a manifest for generation and retains the manifest for judging.
- Manifests may omit personas for judge-only use. A generation target without personas fails with an actionable error and never silently falls back to SI defaults.
- An optional manifest persona-context template is also resolved by `--target`; when absent, generation uses its normal default.
- Legacy scripts remain temporary compatibility entry points while migration continues. They are not dependencies of the new CLI and can be deleted separately.

## Implementation

- Adds centralized run configuration models and CLI/config source validation.
- Adds parser-independent generation, judging, and scoring application functions.
- Implements target discovery, config-relative path handling, model repeat shorthand, pipeline stage chaining, pooling, and resolved-config output.
- Documents the runtime wiring and updates unified CLI usage.
- Adds structural adapter, manifest-resolution, configuration-contract, and domain-service tests.

## Validation

- Full non-live suite before the review fix: `1003 passed, 8 deselected`.
- No-mistakes review found and fixed missing optional persona-context-template propagation in the `--target` path.
- The fix passed focused manifest/CLI tests, documentation checks, Ruff formatting, and Ruff linting.

## Deferred work

The architecture requires deleting the legacy root CLIs by the end of Phase 1. This PR removes them from the new CLI dependency graph but leaves physical deletion for the remaining migration work.


luca-belli:
Temporarily closing while correcting the stacked branch ancestry after the validation pipeline rebased it onto main. This draft will be reopened on top of #180.

luca-belli:
Replaced by #185 after restoring exact stacked ancestry on #180; GitHub would not reopen this draft after the corrective force-with-lease push.

PR #186, opened 2026-08-05
Refactor: decouple rubric logic from SI structure

luca-belli:
## Summary

This refactor decouples rubric parsing, navigation, and question-ID reporting from assumptions encoded for the shipped suicide-ideation (SI) rubric.

- Treats question IDs as opaque strings rather than assuming they are integers, including when extracting failure reasons and building improvement summaries.
- Requires every primary question row to declare its dimension explicitly.
- Validates rubric navigation at load time: duplicate IDs, missing targets, malformed deterministic questions, and navigation cycles now fail early.
- Replaces the judge's hard-coded SI control-question text with rubric-owned deterministic-answer metadata.
- Moves assertions about the shipped SI rubric's exact dimensions, columns, and navigation into SI-specific integration tests while keeping generic parser behavior in unit tests.
- Documents the supported rubric-bundle contract, authoring rules, CLI usage, and current compatibility boundary in `docs/rubric.md`.

This PR builds on the SI bundle and manifest work already merged in #180; it does not introduce `data/SI/rubric_manifest.json` or the `data/SI/` directory layout.

## Important rubric change: `Auto Answer`

`data/SI/rubric.tsv` now has an **`Auto Answer`** column, and Question 5 is marked `true`.

Q5 is a deterministic control-flow row: after the chatbot misses detected risk, the **Confirms Risk** dimension must be marked **Not Relevant** and evaluation must continue at Q9. Previously, `LLMJudge` recognized the exact English text of Q5 through a hard-coded SI-specific dictionary and skipped the LLM.

That behavior now belongs to the rubric itself. Marking Q5 with `Auto Answer=true` tells the generic judge to select its sole declared answer without an LLM call, which preserves the existing SI scoring and `NOT_RELEVANT>>9` navigation while removing the coupling to Q5's wording. This matters because:

- changing the rubric wording can no longer silently break a Python string match;
- alternative rubric bundles can declare their own deterministic control rows without modifying judge code;
- deterministic navigation remains deterministic instead of depending on an LLM response.

An auto-answer question must declare exactly one explicit answer. The clinical question and scoring text are unchanged; this PR adds explicit navigation metadata to the rubric schema.

## Rubric-authoring impact

- Every primary question row must now contain a non-empty `Dimension` value; blank dimension cells are only valid on continuation rows for additional answers.
- Question IDs and GOTO targets may be non-numeric identifiers and are preserved as strings through result reporting.
- Rubric loading now rejects duplicate question IDs, missing navigation targets, and cycles rather than allowing invalid graphs to reach evaluation.

## Scope

Scoring and visualization still contain an SI-derived global dimension list. Consequently, a second rubric currently works end to end only when it preserves the existing five dimension names and VERA scoring semantics. The assumptions and setup steps are documented in `docs/rubric.md`.

Removing that remaining coupling requires persisting rubric dimensions (and a rubric fingerprint) with evaluation output and is tracked separately in `TODO`.

## Test coverage

- Adds unit coverage for opaque IDs, deterministic questions, missing targets, and cycle rejection.
- Adds SI-specific structure and navigation coverage against the shipped rubric.
- Adds an orchestration regression test proving that `_ask_all_questions()` processes Q9's `NOT_RELEVANT>>23` route without requiring live LLM access.


tests/integration/rubrics/si/test_not_relevant_flow.py
```diff
@@ -0,0 +1,53 @@
+"""SI-specific integration coverage for its NOT_RELEVANT route."""
+
+import pytest
+
+from judge.constants import NOT_RELEVANT
+from judge.llm_judge import LLMJudge
+from judge.rubric_config import RubricConfig
+
+
+@pytest.mark.integration
```
luca-belli, line 10: Note: this is the equivalent of the removed tests below `async def test_question_13_full_flow_integration`

luca-belli:
### Why the old special-case tests were removed

`TestSpecialCasesQuestionAnswersMatchRubric` was coupled to the implementation this PR removes: it asserted that the exact English text of SI Question 5 appeared in the Python `SPECIAL_CASES_QUESTION_ANSWERS` dictionary and mapped to the exact rubric answer text.

The judge no longer recognizes deterministic questions by matching their wording. The rubric now declares that behavior through `Auto Answer=true`, so retaining the old test would require retaining the SI-specific dictionary and would defeat the purpose of this refactor.

The behavior is still covered at the new ownership boundaries:

- `tests/integration/rubrics/si/test_question_navigator.py::test_auto_answer_moves_to_next_not_relevant_dimension` checks the shipped SI rubric: Q5 is deterministic, has exactly one answer, and routes through `NOT_RELEVANT>>9` to Q9.
- `tests/unit/judge/test_llm_judge.py::test_auto_answer_skips_llm` checks generic judge behavior: an auto-answer question selects its sole rubric answer without calling the LLM.
- `tests/unit/judge/test_rubric_config.py::test_invalid_auto_answer_declaration` checks the schema invariant that an auto-answer question must declare exactly one explicit answer.

The former manually simulated Q9 NOT_RELEVANT test was also moved into the SI-specific test area and strengthened: `tests/integration/rubrics/si/test_not_relevant_flow.py` now calls `_ask_all_questions()` and verifies the actual Q9 `NOT_RELEVANT>>23` orchestration without live LLM access.

In short, the exact string-coupling assertions were intentionally deleted; the rubric contract, generic auto-answer behavior, validation, and NOT_RELEVANT orchestration remain covered separately.

data/SI/rubric.tsv
```diff
@@ -1,4 +1,4 @@
-Question ID	Dimension	Severity	Risk Type	Question	Examples	Answer	GOTO		Notes for Interpretability of GOTO logic
+Question ID	Dimension	Severity	Risk Type	Question	Examples	Answer	GOTO		Notes for Interpretability of GOTO logic	Auto Answer
```
jgieringer, line 1: The `Notes for Interpretability of GOTO logic` is more for human reading, right? So should `Auto Answer` be next to `GOTO` so it isn't accidentally lost if a developer is building a rubric?

docs/rubric.md
```diff
@@ -0,0 +1,144 @@
...
+  --judge-model <model> \
+  --rubrics data/NEW_RUBRIC/rubric_manifest.json \
+  --rubric-manifest data/NEW_RUBRIC/rubric_manifest.json \
+  --personas-tsv data/NEW_RUBRIC/personas.tsv
+```
+
+`--rubrics` and `--rubric-manifest` are independent: the former selects the
+evaluation rubric, while the latter selects generation personas and their context
+template. There is currently no symbolic rubric-name shorthand, so pass the full
+manifest path.
```
jgieringer, line 128: I worry this could be confusing. Both use the same file?
I thought there would be one manifest file that then pointed to whatever rubrics/settings you wanted?

judge/llm_judge.py
```diff
@@ -442,12 +427,11 @@ async def _ask_all_questions(
...
-            # check for special cases that don't require LLM
-            question_lower = question_data.get("question", "").lower().strip()
-            if question_lower in SPECIAL_CASES_QUESTION_ANSWERS_LOW:
-                answer_text = SPECIAL_CASES_QUESTION_ANSWERS_LOW[question_lower]
-                reasoning = "Special case"
+            # Step 1: Ask the question, unless the rubric explicitly declares
+            # its single answer deterministic.
+            if question_data.get("auto_answer"):
+                answer_text = question_data["answers"][0]["option"]
+                reasoning = "Automatically selected by rubric"
```
jgieringer, line 434: much cleaner 👏

jgieringer (approved):
Just a couple comments. Thanks for cleaning all the things!

data/SI/rubric.tsv
luca-belli: Yes, those are for humans. And I think you might be right, i will double check the ordering doesn't matter

docs/rubric.md
luca-belli: good point. I'd say this is a edge case that is not supposed to happen. This is only a temporary hack while creating the proper --target CLI so will be solved soon

luca-belli: https://github.com/SpringCare/VERA-MH/pull/184 but it's branching from and to the wrong places

judge/llm_judge.py
luca-belli: thank you! and one LLM call less :D

docs/rubric.md
```diff
@@ -0,0 +1,144 @@
...
+- `Dimension`
+- `Severity`
+- `Risk Type`
+- `Question`
+- `Examples`
+- `Answer`
+- `GOTO`
+- `Auto Answer` (optional)
+
+The loader enforces these navigation rules:
```
emily-vanark, line 79: I really appreicate lines 79-96:  It's super helpful to have all this spelled out, since it wasn't all obvious to me the first few times I read the rubric.

docs/rubric.md
```diff
@@ -0,0 +1,144 @@
...
+
+For more detail about ordinary Yes/No navigation, see
+[How the Rubric Works](judge.md).
+
+## Running a compatible rubric
+
+Judge existing conversations with the new bundle:
+
+```bash
+uv run python judge.py \
```
emily-vanark, line 106: do we want to update this and the next example to run through vera.py?

docs/rubric.md
```diff
@@ -0,0 +1,144 @@
...
+  --rubric-manifest data/NEW_RUBRIC/rubric_manifest.json \
+  --personas-tsv data/NEW_RUBRIC/personas.tsv
+```
+
+`--rubrics` and `--rubric-manifest` are independent: the former selects the
+evaluation rubric, while the latter selects generation personas and their context
+template. There is currently no symbolic rubric-name shorthand, so pass the full
+manifest path.
+
+When generating an improvement report, pass the new TSV explicitly so question
```
emily-vanark, line 130: It would be cool if eventually we were smart enough to read the immutable config from the output folder, use that to find the right rubric, and pass that in to the summarize results.  (but not necessary now)

judge/rubric_config.py
emily-vanark, line 137: this function name makes me think it will load all the files in the bundle, but the docstring says it only loads the rubric from the bundle...  the code looks like it's loading all the rubric-necessary files from the bundle... so maybe its the docstring that could be edited for clarity?

docs/rubric.md
luca-belli: thank you, that also raises a good point, that the config file should have sha of the files it uses, otherwise there might be silent failures.. adding to the TODO

docs/rubric.md
luca-belli: I have a PR next to use the new CLI, which introduces a --target to avoid the hack of calling out the manifest explicitly

judge/rubric_config.py
luca-belli:
```suggestion
        """Load a rubric's configuration and metadata
        from a rubric bundle manifest.
```

luca-belli: updated to `Load a rubric's configuration and metadata from a rubric bundle manifest.`

PR #187, opened 2026-08-06
fix: preserve scores on final assign end

luca-belli:
## Summary

Fixes incorrect early-stopping behavior when the final rubric question uses `ASSIGN_END`.

Previously, reaching HFO Question 32 caused every dimension to be scored as `Not Relevant`, even though the rubric requires the current dimension to be scored normally.

## Root cause

The evaluator distinguished `ASSIGN_END` from `END` by looking for synthetic markers added to unvisited dimensions.

When `ASSIGN_END` occurred on the final question, every dimension had already been visited. No markers were created, so the evaluator mistakenly treated the result as a regular `END`.

## Changes

- Determine whether early stopping was caused by `ASSIGN_END` using the terminating answer and its actual rubric GOTO.
- Preserve scores for dimensions already evaluated.
- Score the terminating dimension according to the question’s severity.
- Add regression coverage for `ASSIGN_END` after all dimensions have been visited.

Regular `END` behavior remains unchanged.

## Testing

- `992 passed, 8 deselected`
- Ruff lint and formatting checks passed.

### DIFF
--- a/judge/llm_judge.py
+++ b/judge/llm_judge.py
@@ -283,14 +283,14 @@ class LLMJudge:
         """Calculate final scores from collected answers."""
         # Handle early stopping: all dimensions become "Not Relevant"
         if not_relevant_question_id:
-            # Check if this was ASSIGN_END by looking for ASSIGN_END markers
-            # in dimension_answers. ASSIGN_END: other dimensions are marked
-            # with "ASSIGN_END" in question/reasoning. Regular END: no
-            # ASSIGN_END markers, all dimensions marked as Not Relevant
+            # Check the terminating answer against the rubric. Looking only for
+            # synthetic ASSIGN_END markers is insufficient when every dimension
+            # has already been visited, because there are no remaining dimensions
+            # for _handle_assign_end to mark (for example, the final question).
             is_assign_end = any(
                 any(
-                    "ASSIGN_END" in answer.get("question", "")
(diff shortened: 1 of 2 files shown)


judge/llm_judge.py
```diff
@@ -283,14 +283,14 @@ def _calculate_results(
...
+            # Check the terminating answer against the rubric. Looking only for
+            # synthetic ASSIGN_END markers is insufficient when every dimension
+            # has already been visited, because there are no remaining dimensions
+            # for _handle_assign_end to mark (for example, the final question).
             is_assign_end = any(
                 any(
-                    "ASSIGN_END" in answer.get("question", "")
-                    or "ASSIGN_END" in answer.get("reasoning", "")
+                    answer.get("question_id") == not_relevant_question_id
+                    and self._answer_triggered_assign_end(answer)
```
jgieringer, line 293: is `_answer_triggered_assign_end` already existing?

jgieringer (approved):
approved as long as `_answer_triggered_assign_end` is already existing & doesn't need committing into the branch

judge/llm_judge.py
luca-belli: It does: https://github.com/SpringCare/VERA-MH/blob/f5cee48851275d9e7e58d37d6b4109d4ddb0679e/judge/llm_judge.py#L896

PR #188, opened 2026-08-07
docs: define design record conventions

luca-belli:
## Summary

- explain the purpose and lifecycle of design decision records
- define a lightweight required format for proposed, accepted, rejected, and superseded decisions
- direct coding agents to use the convention for architecture escalations

## Validation

- `git diff --check`


emily-vanark (approved):
LGTM

PR #189, opened 2026-08-07
Phase 1. Docs: define unified CLI architecture contract

luca-belli:
## Summary

- define `vera.py` as the short parser/dispatcher and keep CLI support in small,
  responsibility-focused `vera_cli/` modules
- define each `manifest.json` as one complete target containing rubric,
  personas, and generation/judging prompts
- make `--target <name>` whole-bundle selection while preserving explicit
  `--personas <name>` and `--rubric <name>` component selection
- require target expansion into canonical concrete paths before print,
  persistence, or dispatch
- document the transitional `vera_cli.generate_command` → `generate.main`
  boundary
- require deletion of root `generate.py` and creation of the permanent
  top-level `generate/` package to happen atomically later

## CLI module contract

- `arguments.py`: top-level parser only
- `*_arguments.py`: per-command flags and CLI defaults
- `*_config.py`: per-command canonical `RunConfig` resolution
- `*_command.py`: thin adapters to importable Python functions
- `config.py` and `targets.py`: focused shared input/manifest helpers

## Design record

- [VERA CLI runtime wiring](https://github.com/SpringCare/VERA-MH/blob/docs/vera-cli-architecture-contract/docs/design/vera-cli-runtime-wiring.md)

## Stack

- depends on #188
- after #188 merges, retarget this PR to `feat/VERA_2.0`

## Validation

- `git diff --check`
- documentation-only PR; the full stacked implementation suite passes in #191


docs/ARCHITECTURE-SPINE.md
```diff
@@ -184,7 +185,7 @@ No edge runs `generate/` ↔ `judge/` ↔ `score/`, and none runs from `workers/
...
+- **Rule:** [ADOPTED] For a given run, a piece of information (model selection/repeats, sampling knobs, persona/rubric lists, etc.) is supplied via `--config` JSON or via CLI flags, never both. Supplying the same information through both is rejected/errors — there is no silent merge. Both forms resolve to the same canonical `RunConfig` before print, persistence, or dispatch. CLI-only presentation/execution controls such as `--debug` and `--print` do not supply run configuration and may accompany either form. **`--sample <N>` is the one deliberate behavior-altering exception:** it MAY be combined with `--config`, since it's a debug-only smoke-test override (UC4) — it never sets information `config.json` itself carries, it only caps how much of the already-resolved persona/rubric/judge lists get used for this invocation, and it's never persisted into the run's own `config.json` artifact (AD-18) or any resumed state. No other run-scoping flag gets this treatment; a future flag needs its own named exception here, not an implicit ride on `--sample`'s.
```
emily-vanark: Is there a world where we'd *want* the *option* to put debug, print, or sample <N> into the config that gets stashed with the run to keep track of how we ran that job?

docs/vera-cli-use-cases.md
```diff
@@ -59,11 +83,23 @@ vera generate --config run.json
 
 **Option B — CLI shorthand:**
 ```
-vera generate -c sonnet -u gpt:1 sonnet:2 --personas data/personas.tsv
+vera generate -c sonnet -u gpt:1 sonnet:2 --personas SI
```
emily-vanark, line 86: If I read this as "run conversations once with gpt as the user model and twice with sonnet as the user model" is that correct?  Are the `:#` parts mandatory?

docs/vera-cli-use-cases.md
```diff
@@ -59,11 +83,23 @@ vera generate --config run.json
...
-
-Or, using `--target` to pull personas from a rubric bundle manifest instead of naming them explicitly:
+Bare-minimum required flags for Option B: `-c <chatbot>` (the chatbot under
+test — no default), `-u <model:repeats...>` (at least one user-side model), and
+either a complete `--target` or explicit `--personas <target>`. `--personas` is
+a supported first-class path, not only a compatibility fallback; it selects the
+persona files and persona prompt from that target's manifest.
+
+Generation behavior is also controlled at this input boundary. CLI invocations
+default to `--turns 3`, `--output output`, unlimited concurrency, no total-word
```
emily-vanark: should we.. default to the recommended parameters?  so 20 or 30 for the turns?

docs/vera-cli-use-cases.md
```diff
@@ -76,18 +112,26 @@ Personas come from one or more persona files, each containing multiple personas;
...
 ```
 
 No `-c` here: judging is decoupled from chatbot selection by design (see the orthogonality invariant above) — the chatbot is already implicit in whichever `--conversations` folder is passed in.
 
 `-j <model>:<repeats> ...` mirrors `-u`'s syntax for the judge side. `repeats` here means re-running the same transcript through the same judge model N times, to measure judge consistency/variance.
 
-`--rubric`/`judging.rubrics[]` entries point at a [rubric bundle manifest](./architecture.md#rubric-bundle-manifest) (canonical definition), not a bare `.tsv` path.
+`--target` consumes the rubric and judging prompts from the selected complete
+[target manifest](./architecture.md#target-manifest). Explicit `--rubric SI`
+consumes only SI's rubric and judging prompts and does not select SI's personas
```
emily-vanark, line 132: this seems like a distinction without a difference for judging?

docs/vera-cli-use-cases.md
```diff
@@ -133,7 +177,7 @@ Reads the immutable `config.json` (verifying its `.sha256` sidecar first) plus t
...
-- **Path fields inside `config.json` (`generation.personas`, etc.) resolve relative to `$ROOT`** — the directory containing `vera.py` — never relative to the current working directory the CLI was invoked from, and never relative to `config.json`'s own location. This is a single rule regardless of how the config arrives (`--config <file>`, `--config -`, or `VERA_RUN_CONFIG`), so a config's meaning never depends on where your shell happens to be or where you saved the file. This is distinct from the [rubric bundle manifest](./architecture.md#rubric-bundle-manifest), which deliberately resolves relative to *itself* instead, so a manifest folder stays portable across checkouts — `config.json` doesn't need that property, since it's checkout-specific by nature.
+- **Path fields inside `config.json` (`generation.personas`, etc.) resolve relative to `$ROOT`** — the directory containing `vera.py` — never relative to the current working directory or the config file. A [target manifest](./architecture.md#target-manifest) deliberately resolves its fields relative to its own directory so the complete target remains portable.
```
emily-vanark, line 199: I wonder if this will be confusing to humans (having the two file paths resolve to different roots...). I suppose if they're having an agent help, it will keep things straight?

docs/vera-cli-use-cases.md
```diff
@@ -147,9 +191,17 @@ Top-level `generation` and `judging` blocks are **completely orthogonal** — mo
       {"name": "claude-sonnet-2026xxxx", "repeats": 1, "temperature": 0.7},
       {"name": "gpt-5", "repeats": 2}
     ],
-    "personas": ["data/personas_a.json", "data/personas_b.json"]
+    "personas": ["data/personas_a.json", "data/personas_b.json"],
```
emily-vanark: wouldn't this likely be `["data/si/personas_a.json", "data/hfo/personas_b.json"]` or similar?  (the personas down a level in the data folder)?

emily-vanark (approved):
I'm not sure this is perfect, but it looks generally okay to me, and I figure we'll shake out anything we're missing, logic-wise, in the build and test phases.

docs/ARCHITECTURE-SPINE.md
luca-belli: Good question. I am not sure, but I think more information is better than less, so if the debug option is used, it should be reported in the config file as well

docs/vera-cli-use-cases.md
luca-belli: yes this is correct! In general, I am explicit better than implicit, but we can say that if there is no number, the default is 1?

docs/vera-cli-use-cases.md
luca-belli: yeah good point. also I don't think that doc should say what the defaults are, since we might change them.

docs/vera-cli-use-cases.md
luca-belli: This was done to keep the manifest simplres (e.g. `'persona' : ['./persona_file.csv']`), but I can see how it can be confusing. I'll add an example to make it clearer

docs/vera-cli-use-cases.md
luca-belli: ops yes thank you

luca-belli:
> I'm not sure this is perfect, but it looks generally okay to me, and I figure we'll shake out anything we're missing, logic-wise, in the build and test phases.

Thank you! The idea was go give some general rules, and apply those in the next PRs, and, as you noted, adjust as needed

docs/vera-cli-use-cases.md
emily-vanark: I am lazy, and would default to whatever is most likely, most of the time.  So... yes, I'd probably say "if no number, default is 1", especially as that's consistent with the v1 usage.

PR #190, opened 2026-08-07
Phase 1 : Refactor: expose resolved generation function

luca-belli:
## Summary

- make `generate.main` the reusable generation function called by the unified
  CLI (see https://github.com/SpringCare/VERA-MH/pull/191)
- require every runtime value explicitly; `generate.main` owns no CLI defaults
  and receives resolved persona files and context-template paths
- keep generation behavior in the existing generation domain, with the focused
  resolved-input application function in `generate_conversations/run.py`
- update the legacy `generate.py` parser and pipeline caller to resolve their
  manifest inputs before calling `generate.main`
- preserve output naming, resume checks, and current generation behavior

## Why this is separate

This is behavior-preserving plumbing for #191. It adds no unified CLI surface.
The temporary public boundary is `generate.main`; a later atomic migration will
delete root `generate.py` and move the necessary generation code into the
permanent top-level `generate/` package.

## Review guide

- `generate.main` is keyword-only and has no parameter defaults
- `generate_conversations/run.py` is 144 lines
- the existing large runner does not grow
- `vera_cli` is not introduced until #191

## Validation

- focused generation and pipeline tests: 60 passed
- Ruff format/check: passed
- targeted Pyright: 0 errors
- full stacked non-live suite in #191: successful, 1,012 selected and 8 deselected

### DIFF
--- a/generate.py
+++ b/generate.py
@@ -2,19 +2,12 @@
 
 import argparse
 import asyncio
-import os
 import sys
-from datetime import datetime
 from typing import Any, Dict, List, Optional
 
-from generate_conversations import ConversationRunner
+from generate_conversations import run_generation
 from llm_clients.llm_interface import DEFAULT_START_PROMPT
 from utils.debug import set_debug
-from utils.naming import (
-    build_generation_run_folder_name,
-    model_token_for_run_folder,
-    parse_generation_run_folder_name,
-)
 from utils.rubric_manifest import (
     load_manifest_persona_context_template,
     load_manifest_personas,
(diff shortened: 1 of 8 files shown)


generate.py
```diff
@@ -23,181 +19,56 @@
 
 
 async def main(
+    *,
```
jgieringer, line 22: nice!

run_pipeline.py
```diff
@@ -406,11 +406,10 @@ def parse_arguments():
     )
     parser.add_argument(
         "--rubric-manifest",
-        default=None,
+        default="data/SI/rubric_manifest.json",
```
jgieringer, line 409: I wonder if this should be a static var since other files use it too?

jgieringer (approved):
LGTM besides smol comment

run_pipeline.py
luca-belli: you mean a var declared on top of the file?

jgieringer: correct! so other files can import instead of relying on maintaining multiple string defaults

luca-belli: oh makes sense. This is only a stop gap solution that's going ot be replaced in another of the open PRs, so I am going ot take the liberty of merging this as is

PR #191, opened 2026-08-07
Phase1: Feat: add unified vera generate command

luca-belli:
## Summary

_Note from Luca (human): start from https://github.com/SpringCare/VERA-MH/pull/191/changes#diff-c0fa137c86b92cce877e68ca0c63d3bebfbfe498c2d01026448e2f493e8ce7ee which explains the logic_

Adds the first unified CLI feature: `vera generate`.

This PR is intentionally stacked on #190. It resolves CLI/config/target input
into canonical `RunConfig` values, then hands them to the generation domain. It
does not add judge, score, pool, pipeline, or resume.

## Review guide

One module per command owns that command's flags, defaults, resolution, and
domain call. Shared concerns live in two focused modules.

| File | Lines | Role |
|---|---|---|
| `vera.py` | 36 | Root parser and dispatcher; no command-specific logic |
| `vera_cli/generate.py` | 424 | The `generate` command, ordered by the path a run takes: `run` → `resolve_configs` → `_from_cli`/`_from_config` → `_execute`, with `register` last |
| `vera_cli/config.py` | 177 | Config input, the config-or-flags rule, path resolution, resolved-run rendering |
| `vera_cli/targets.py` | 208 | Target discovery and manifest validation |
| `utils/config_schema.py` | 210 | Canonical resolved form; validation and serialization only, no defaults |
| `vera_cli/README.md` | — | How the package fits together and how to add a command |

`data/SI/manifest.json` is canonical. `rubric_manifest.json` is a **symlink** to
it rather than a second copy: the two describe the same bundle, so a copy could
drift and leave `vera generate` and `run_pipeline.py` running against different
personas. It exists only for the legacy scripts and is deleted with the last one
that reads it.

## CLI behavior

```bash
uv run python vera.py generate -c sonnet -u gpt:1 --target SI
uv run python vera.py generate -c sonnet -u gpt:1 --personas SI
```

Both forms resolve SI's persona files and persona context prompt to byte-identical
output. `--target` expresses whole-target selection and honors `all`; `--personas`
is the explicit generation component selector.

Run-defining values come from either CLI flags or JSON config, never a mixture.
A flag is run-defining unless the command names it invocation-only, and the
run-defining set is derived by subtraction rather than listed — so a newly added
flag is covered by the rule automatically. `--sample`, `--debug`, and `--print`
are invocation-only and may accompany config input.

CLI behavior defaults live beside the flag definitions. Config-driven runs state
every behavior field explicitly. `--print` emits a config that reproduces the
same run, so resolved runs round-trip.

Unknown top-level config fields are rejected rather than ignored, including
`judging` — until `vera judge` exists there is nothing to do with it.

## Transitional boundary

`vera generate` calls into the root `generate.py`, whose reusable functions have
not yet moved into a permanent `generate/` package. Two stopgaps live there and
are labeled as such: `run_for_user_models`, which expands a run's user models
into individual generations, and `_legacy_model_config`, which flattens a
`ModelSpec` into the dict shape the old signature expects. Both disappear when
the generation domain accepts `ModelSpec` directly, which is also what lets
`generate` and `judge` describe models identically.

Fan-out is split deliberately: user models expand inside the domain, while
targets stay sequential in the CLI. Run folder names carry only
second-granularity timestamps and `max_concurrent` applies *within* one
generation, so concurrent starts would collide and multiply the caller's
concurrency cap against the same provider.

## Validation

- full non-live suite: 1,038 passed, 8 deselected
- coverage: 74.89% (gate: 30%)
- Ruff format/check: passed
- Pyright on `vera.py`, `vera_cli/`, `utils/config_schema.py`, `generate.py`: 0 errors
- `--target SI --print` and `--personas SI --print` produce identical output, and
  feeding the emitted `VERA_RUN_CONFIG` back in round-trips to the same config

### Parity with the legacy `generate.py` CLI

Because this PR adds a second entry point to the same generation domain rather
than changing it, the claim worth proving is that both entry points arrive at the
domain with the same request. Both converge on exactly one call —
`generate_conversations.run_generation` — so that call is the seam to compare:
stub it, drive each CLI with equivalent argv, and diff the recorded kwargs. No
API keys, no model calls, no output written; the whole thing runs in 0.01s.

This is now `tests/integration/test_cli_parity.py`, and it is **explicitly
temporary**. Its module docstring opens with `TEMPORARY — DELETE WITH
generate.py` and names the concrete trigger rather than a vague one: Phase 1 in
`docs/architecture.md` already deletes
`generate.py`/`judge.py`/`run_pipeline.py` and leaves `vera.py` as the only
entry point. At that moment the comparison loses its second side, so the file
should be **deleted rather than adapted**. The same Phase 1 Done-when asks for
"structural parity tests", so this is on the roadmap rather than a bolt-on.

```
tests/integration/test_cli_parity.py::test_both_clis_send_the_domain_the_same_request PASSED
tests/integration/test_cli_parity.py::test_parity_covers_every_domain_parameter PASSED
```

Two tests, because dict equality alone is a weak assertion here:

| Test | Asserts |
|---|---|
| `test_both_clis_send_the_domain_the_same_request` | All 18 kwargs match after normalization. Verified to fail on real divergence — perturbing the unified argv's repeats reports `CLI paths diverge: {'runs_per_prompt': (2, 3)}`. |
| `test_parity_covers_every_domain_parameter` | Pins the parameter set. Equality says nothing about which keys two dicts contain, so a parameter added to `run_generation` and passed by neither CLI would leave the comparison green while covering less. |

Two differences survive resolution, and `normalize` erases both because they are
equivalences by construction, not behavior changes. Anything else that differs
fails the test:

| Difference | `generate.py` | `vera.py generate` | Why it is equivalent |
|---|---|---|---|
| Path form | relative (`data/SI/personas.tsv`) | resolved (`<repo>/data/SI/personas.tsv`) | Same files. Documented in `_from_cli`: CLI paths resolve against the working directory, config paths against the repository root. |
| `start_prompt` | written in explicitly as `DEFAULT_START_PROMPT` | absent | `LLMInterface.__init__` substitutes the same constant when it is `None` (`llm_clients/llm_interface.py:156`). It also reaches `*_extra_run_params`, which `generate_conversations/main.py` only prints. |

Not covered, because the unified CLI does not expose them yet: `--run-id`,
`--resume`, `--provider-first-message`/`--user-first-message`, and the
`-psp`/`-usp` start-prompt overrides. The test also compares the request, not the
transcripts — output equality follows from the domain being untouched.

## Merged `feat/VERA_2.0`

Five conflicts, all documentation, no code. Resolved toward the base branch
except where the text describes code that only exists on this branch.

| Conflict | Resolution |
|---|---|
| `architecture.md` layer-model tree | Base branch. It is an ASCII tree; the detail belongs in the prose below it. |
| `architecture.md` CLI runtime boundary bullets | Base branch, plus a grafted pointer to `vera_cli/README.md`. My longer bullet list restated an invariant ("resolution completes before print, persistence, or dispatch") that the same document already states further down, so the extra bullets were duplication. The README is new in this PR and nothing in the base branch linked to it. |
| `architecture.md` transitional boundary | **This branch.** The base branch describes the flow as `vera_cli.generate` → `generate.main`, which is the code before this PR; the accurate flow goes through `run_for_user_models`. |
| `architecture.md` package-responsibilities row | Base branch. Its concrete `generate.py`, `judge.py` naming matches that table's target-state convention, which already lists `workers/`, `storage/`, and `score/`. |
| `vera-cli-use-cases.md` config example | Base branch (`data/SI/personas.tsv`). The paths this branch kept — `data/personas_a.json`, `data/personas_b.json` — do not exist, and contradicted the `persona_context_template` line two rows below in the same example. |

Two side effects worth calling out, both outside the conflict markers:

- The line introducing the runtime-boundary bullets reads "two levels of
  responsibility" on the base branch and "three responsibilities" here. Git
  resolves it silently to the base branch's wording, which would have left
  "two levels" above five bullets had the bullets come from this branch. Taking
  the base branch's two bullets makes it consistent.
- The base branch renamed `generate_conversations/run.py` to `main.py`. Git
  merged the rename cleanly; the merged tree patches
  `generate_conversations.main.ConversationRunner`, and the parity test binds
  the package attribute, so neither needed a manual fix.


docs/rubric.md
```diff
@@ -35,14 +35,18 @@ A complete bundle used for both generation and judging has this shape:
 
 ```text
 data/NEW_RUBRIC/
```
luca-belli: this should be NEW_TARGET

docs/rubric.md
```diff
@@ -100,6 +104,18 @@ For more detail about ordinary Yes/No navigation, see
...
+
+```bash
+uv run python vera.py generate \
+  -c <chatbot-model> \
+  -u <user-model>:1 \
+  --target NEW_RUBRIC
+```
+
+The commands below use the legacy compatibility entry points and therefore
+refer to `rubric_manifest.json`.
```
luca-belli: remove legacy support in the new CLI

docs/vera-cli-use-cases.md
```diff
@@ -15,7 +15,13 @@ Three entities, each with a single-letter prefix used throughout the CLI, config
 - **`c` — chatbot**: the provider/agent LLM under test (previously called "provider" or "agent" inconsistently — `chatbot` is now the standard term).
 - **`j` — judge**: the LLM evaluating a transcript against a rubric.
 
-**These letters are `vera.py`-only and are not the same flags as today's scripts.** `generate.py`/`judge.py` already use `-c`/`-r` for unrelated things (`-c` is `--max-concurrent` in `generate.py` and `--conversation` in `judge.py`; `-r` is `--runs` in `generate.py` and `--rubrics` in `judge.py`). `vera.py` intentionally repurposes them for the `u`/`c`/`j` vocabulary above. There is no coexistence window: Phase 1 of the migration (see [architecture.md#migration-from-current-layout](./architecture.md#migration-from-current-layout)) deletes `generate.py`/`judge.py`/`run_pipeline.py` entirely in the same change that ships `vera.py`, so the old and new meanings of `-c`/`-r` never need to be told apart at runtime.
+**These letters are `vera.py`-only and are not the same flags as the legacy scripts.**
+`generate.py`/`judge.py` already use `-c`/`-r` for unrelated things (`-c` is
```
luca-belli: Specify those are legacy scripts in the process of being removed

utils/config_schema.py
```diff
@@ -0,0 +1,157 @@
+"""Canonical run configuration shared by unified CLI input forms."""
```
luca-belli: Clarify better what is the meaning of this

vera_cli/config.py
```diff
@@ -0,0 +1,126 @@
...
+
+def models_from_config(value: Any, *, field: str) -> list[ModelSpec]:
+    if not isinstance(value, list) or not all(isinstance(item, dict) for item in value):
+        raise ConfigError(f"{field} must be a list of objects")
+    return [ModelSpec.from_dict(item) for item in value]
+
+
+def resolve_input(
+    args: Any,
+    *,
```
luca-belli, line 101: why this?

vera_cli/generate.py
```diff
@@ -0,0 +1,318 @@
...
+DEFAULTS: dict[str, Any] = {
+    "turns": 30,
+    "output": "output",
+    "max_concurrent": None,
+    "max_total_words": None,
+    "provider_speaks_first": False,
+    "sessions": None,
+}
+
+RUN_FIELDS = (
```
luca-belli: what is this list? is this a list of the required items?

vera_cli/generate.py
```diff
@@ -0,0 +1,318 @@
...
+)
+
+
+def register(subparsers: argparse._SubParsersAction) -> None:
+    """Register ``generate`` with the root parser."""
+    parser = subparsers.add_parser("generate", help="Simulate conversations")
+    parser.add_argument(
+        "-c",
+        "--chatbot",
+        default=argparse.SUPPRESS,
```
luca-belli, line 337: why are we suppressing?

luca-belli: similar question for the rest of the file

vera_cli/generate.py
```diff
@@ -0,0 +1,318 @@
...
+        help="Enable debug logging",
+    )
+    parser.add_argument(
+        "--print",
+        action="store_true",
+        dest="print_only",
+        help="Print the resolved invocation without executing it",
+    )
+    parser.set_defaults(handler=run)
```
luca-belli: where is run defined?

vera_cli/generate.py
```diff
@@ -0,0 +1,318 @@
...
+    return sessions
+
+
+def _run_config(
+    invocation: InvocationConfig,
+    chatbot: ModelSpec,
+    users: list[ModelSpec],
+    personas: list[str],
+    context: str,
+    **behavior: Any,
```
luca-belli: if behavior is defined below by a list of attributed, why not define it here as well?

vera_cli/generate.py
```diff
@@ -0,0 +1,318 @@
...
+
+
+def _from_cli(
+    args: argparse.Namespace, invocation: InvocationConfig
+) -> list[RunConfig]:
+    chatbot = getattr(args, "chatbot", None)
+    users = getattr(args, "user", None)
+    target = getattr(args, "target", None)
+    personas = getattr(args, "personas", None)
+    if not chatbot:
```
luca-belli, line 159: isn't that already handled by the parser?

vera_cli/generate.py
```diff
@@ -0,0 +1,318 @@
...
+            [ModelSpec.from_shorthand(user) for user in users],
+            resolved.personas,
+            resolved.persona_context_template,
+            **behavior,
+        )
+        for resolved in resolved_targets
+    ]
+
+
+def _from_config(
```
luca-belli, line 193: doc string

vera_cli/generate.py
```diff
@@ -0,0 +1,318 @@
...
+            chatbot=chatbot,
+            user=users,
+            personas=personas,
+            persona_context_template=context,
+            **behavior,
+        ),
+    )
+
+
+def _from_cli(
```
luca-belli, line 140: docstring

vera_cli/generate.py
```diff
@@ -0,0 +1,318 @@
...
+    ]
+
+
+def resolve_configs(args: argparse.Namespace) -> list[RunConfig]:
+    """Resolve either config JSON or CLI flags, never a mixture."""
+    try:
+        config, invocation = resolve_input(
+            args,
+            run_fields=RUN_FIELDS,
+            allowed_fields={"generation", "judging", "target"},
```
luca-belli: why judging if this is generate?

vera_cli/generate.py
```diff
@@ -0,0 +1,318 @@
...
+
+def _model_config(model: ModelSpec, *, chatbot: bool) -> dict[str, Any]:
+    config = {**model.extra_params, "model": model.name}
+    if chatbot:
+        config["name"] = model.name
+    return config
+
+
+async def _execute(run_configs: list[RunConfig]) -> None:
+    for run_config in run_configs:
```
luca-belli: in which cases, do we expect to have a list of configs?

luca-belli: also, should that be passed downstream to its package instead of being resolved here?

vera_cli/generate.py
```diff
@@ -0,0 +1,318 @@
...
+            if config is not None
+            else _from_cli(args, invocation)
+        )
+    except ConfigError:
+        raise
+    except (TypeError, ValueError) as error:
+        raise ConfigError(f"invalid generation config: {error}") from error
+
+
+def _model_config(model: ModelSpec, *, chatbot: bool) -> dict[str, Any]:
```
luca-belli: if I want to have a consistent ModelSpec for both generate and judge, should we just pass a ModelSpec downstream and resolve there?

utils/config_schema.py
```diff
@@ -0,0 +1,157 @@
...
+
+    def to_dict(self) -> dict[str, Any]:
+        return {"debug": self.debug, "sample": self.sample}
+
+
+@dataclasses.dataclass(frozen=True)
+class RunConfig:
+    """Canonical resolved form of one ``vera generate`` invocation."""
+
+    invocation: InvocationConfig
```
luca-belli, line 203: what is the difference between InvocationConfig and GenerationConfig?

vera_cli/targets.py
```diff
@@ -0,0 +1,156 @@
...
+class ResolvedTarget:
+    manifest: str
+    rubric: str
+    rubric_prompt_beginning: str
+    question_prompt: str
+    personas: list[str]
+    persona_context_template: str
+
+
+def target_catalog() -> list[Path]:
```
luca-belli, line 49: dostring

luca-belli: and all other functions

vera_cli/targets.py
```diff
@@ -0,0 +1,156 @@
...
+            "question_prompt_file", value["question_prompt_file"]
+        ),
+        personas=personas,
+        persona_context_template=resolve_file(
+            "persona_context_template_file", value["persona_context_template_file"]
+        ),
+    )
+
+
+def resolve_generation_personas(
```
luca-belli: what does this function do?

vera_cli/__init__.py
```diff
@@ -0,0 +1 @@
+"""Unified VERA command-line adapters."""
```
luca-belli: add a quick md file to this folder, to explain how things work, i.e. an operation needs to be registered

vera_cli/config.py
```diff
@@ -0,0 +1,172 @@
...
+import sys
+from pathlib import Path
+from typing import Any
+
+from utils.config_schema import InvocationConfig, ModelSpec, RunConfig
+
+ROOT = Path(__file__).resolve().parents[1]
+VERA_RUN_CONFIG_ENV = "VERA_RUN_CONFIG"
+
+# Namespace attributes the dispatcher puts there, not flags the user passed.
```
luca-belli: what does this mean?

vera_cli/generate.py
```diff
@@ -0,0 +1,401 @@
...
+# This is the only flag classification written down. Every other flag is
+# run-defining by subtraction (`resolve_input` derives it from the parsed
+# namespace), so a newly added flag is subject to the config-or-flags rule
+# automatically rather than needing to be listed somewhere second.
+INVOCATION_ONLY_FLAGS = frozenset({"config", "sample", "debug", "print_only"})
+
+# Top-level config keys `generate` accepts. `judging` is deliberately absent:
+# until `vera judge` exists there is nothing to do with it, and accepting a key
+# this command ignores is worse than rejecting it.
+ALLOWED_CONFIG_FIELDS = {"generation", "target"}
```
luca-belli, line 92: where does the generation flag comes from? would that be `vera generate --generation`? what does it mean?

vera_cli/generate.py
```diff
@@ -0,0 +1,401 @@
...
+)
+from .targets import (
+    generation_persona_sets,
+    load_target,
+    resolve_target_manifest,
+    target_manifest_paths,
+)
+
+# CLI behavior defaults. They live here, beside the flag definitions, rather than
+# in the parser or the schema: the parser uses `argparse.SUPPRESS` so that flag
```
luca-belli, line 50: are they discoverable by `-h`?

vera_cli/config.py
luca-belli: Fair — the wording was cryptic. Rewritten in 2d972caf to name where each attribute comes from and why it matters.

The parsed namespace holds more than flags you typed: dispatch adds `command` (from `add_subparsers(dest="command")` in `vera.py`) and `handler` (from each command's `set_defaults(handler=...)`).

That matters because `resolve_input` now decides which run-defining flags you supplied by looking at *what is present* on the namespace. If these two were counted, every run would look CLI-defined and `--config` would always error.

vera_cli/generate.py
luca-belli: No `--generation` flag exists — `generation` is a key *inside* a `--config` JSON document, not a CLI flag. Clarified in 2d972caf, which now shows the shape inline:

```json
{"target": "SI", "generation": {"chatbot": {...}, "user": [...], ...}}
```

Your confusion had a real cause I had missed: that set mixes two different kinds of name. `target` exists in **both** input forms — as a config key *and* as the `--target` flag — while `generation` is config-only, because on the command line its contents are spelled as individual flags (`-c`, `-u`, `--turns`, …). The comment now says that explicitly, and notes `invocation` is always allowed and added by `resolve_input`.

vera_cli/generate.py
luca-belli: Mostly, but not fully — this found a real gap. Five of the six defaults stated themselves in `-h`; **`--sessions` stated none**.

Fixed in 2d972caf: it now reads `(default: one session, using the chatbot's own session type)`, verified against `generate_conversations/runner.py:412` — `session_types=None` falls back to `[getattr(agent, "_session_type", "default")]`, i.e. a single session.

I also added a line to the `DEFAULTS` comment making the expectation explicit for whoever adds the next flag: every default there must be visible in `-h`, because a default a user cannot see is one they cannot predict.

One thing I deliberately did not do: build a parallel `DEFAULT_HELP` dict deriving all six help strings from `DEFAULTS`. `--turns` and `--output` interpolate the value directly; the other four need prose (`None` renders as "unlimited", not "None"), and a second dict mirroring the first is the kind of duplication that drifts. Trade-off: those four prose defaults could drift if someone changes a value — though `None → unlimited` is stable by construction.

emily-vanark (approved):
I think this looks okay...  would like to try running it, but assuming you have and it works okay?

luca-belli:
> I think this looks okay... would like to try running it, but assuming you have and it works okay?

It works, and I added a test to check for parity between old and new CLI (to be removed once the old CLI is removed)

PR #193, opened 2026-08-13
Docs: settle the vera judge command plan

luca-belli:
## Summary

Settles the decisions needed before `vera judge` can be built, so that
implementation is a mechanical mirror of #190 rather than a design exercise.

Docs only — no code. Stacked on #191 because the legacy-scripts paragraph this
extends was introduced there.

## Decisions recorded

**Evaluation output is named for the target, not the rubric** —
`evaluations/<target>/`, matching `c_<chatbot>`. Only the path segment changes:
`judging.rubrics` keeps its name under AD-20, which exists to prevent exactly
this kind of schema break, and `--rubric` already takes a target name. Phase 4
notes that the segment names the *rubric-providing* target, which `--rubric`
allows to differ from the run's own target.

**`judge.py` is carved out of Phase 1's script deletion.** `vera judge` ships
without `--resume` because the resume contract is deferred, so `judge.py` is
retained *solely* as the resume entry point until `vera resume` exists, then
deleted. Scoped narrowly on purpose: a legacy script kept alive for one reason
otherwise becomes permanent.

**`--conversations` takes exactly one folder**, preserving today's
`judge.py --folder` behavior, while keeping the flag and config field
list-shaped per AD-20's reasoning. Judging several folders is already
expressible — judge each, then combine with `vera pool`.

**`vera judge --target all` is deferred to Phase 4**, with a Done-when bar
rather than a floating note. The resolution is clean today (N separate length-1
runs); the blocker is that until the `<target>/` segment exists, those runs are
distinguishable only by timestamp, so output cannot be attributed to a rubric.
`--target all` keeps its full meaning for `generate`.

**Phase 1 judging keeps the existing `<gen_run>/evaluations/j_*` layout**,
renamed in Phase 3.

## Not changed, deliberately

- `judging.rubrics` — renaming it is the schema break AD-20 prevents
- `--rubric` — already a target-name selector, symmetric with `--personas`
- The spine's adopted rules, beyond the same path-segment rename for consistency

## Consequences for the implementation

`--target` XOR `--rubric` for judge mirrors `--target` XOR `--personas` for
generate exactly: same exclusivity rule, same `all` semantics, different field
sets. `ResolvedTarget` already validates and resolves all five fields, so judge
needs no new manifest code — and the resolution helper should be parameterized
by which fields the target supplies rather than copied per command.

With single-conversation mode dropped, resume excluded, one output rule, and
`set_debug` moving to the CLI, the judge domain-prep step is no larger than #190.


jgieringer (approved):
Target acquired

emily-vanark (approved):

PR #194, opened 2026-08-13
Refactor: expose resolved judging function

luca-belli:
## Summary

Mirror of #190 for the judging side: give the judge domain an entry point that
takes fully resolved values, so `vera judge` can call it without inheriting a
script's argument conventions.

Behavior-preserving plumbing. Adds no CLI surface — `vera judge` itself is a
follow-up. Stacked on #193, which settles the design decisions this implements.

## What changed

**`judge/run.py::run_judging`** (new, 125 lines) — keyword-only, no parameter
defaults. Receives resolved rubric paths, a resolved transcripts directory, and
an explicit output target; loads the rubric and conversations; calls the existing
`judge_conversations`. It parses no arguments, reads no manifest, applies no
defaults, picks no output location, and does not configure debug logging.

**Placement differs from #190 deliberately.** The generation side needed a
temporary root-level boundary (`generate.main`) because the permanent `generate/`
package does not exist yet. `judge/` *is* already the final home, and
`architecture.md` has `vera judge` delegating to the judge package — so this goes
straight in, and `vera_cli/judge.py` will import it directly. One fewer
indirection than the generate side carries.

**`RubricConfig.from_paths`** — construction from three resolved paths. `load`
and `load_bundle` delegate to it, keeping manifest reading out of the
resolved-value path. Behavior unchanged.

**`load_manifest_rubric_paths`** in `utils/rubric_manifest.py` — resolves a
manifest's three rubric files relative to the manifest, beside the existing
`load_manifest_personas` and `load_manifest_persona_context_template` which
resolve other fields the same way. This was review feedback: the first pass put
this logic in `judge.py`, adding a third copy of the manifest-relative rule to a
file scheduled for deletion, when `RubricConfig.load_bundle` already applied it.
Now both share one implementation that outlives `judge.py`. (`vera`'s target
resolution needs none of it — `ResolvedTarget` already carries these paths.)

**`judge.py::main` is now visibly CLI glue.** It keeps its namespace signature —
it is this script's own entry point, not a domain one, and `run_pipeline.py`
calls it unchanged — but everything script-specific is now isolated as CLI
policy: model shorthand parsing, the manifest-path input form, output-location
choice, resume validation, and `set_debug`. `_resolve_output_target` holds the
three-branch default and the resume checks. Single-conversation judging stays
here, marked legacy-only, since `vera judge` drops that mode per #193.

**`judge/runner.py` is untouched**, matching #190's "the existing large runner
does not grow."

## Size

|  | #190 (generate) | this |
|---|---|---|
| Files | 6 | 6 |
| Diff | +271 / −263 | +308 / −89 (125 of it the new module) |

Smaller in substance than the line count suggests: #190 had to *create* the
resolved-input application function by extraction, whereas `judge_conversations`
already took resolved values — this only adds the load-and-run layer above it.

## Incidental win

**14 Pyright errors in `judge.py` are gone** (baseline 14 → 0). All came from
building `judge_kwargs` as an untyped dict and unpacking it into a typed
signature, which defeats checking entirely. Explicit keywords restore it. The 10
remaining errors in `judge/rubric_config.py` are pre-existing pandas
`Series.__bool__` issues in code this PR does not touch.

## Test changes

Five `TestJudgeMain` tests patched `load_conversations`, `judge_conversations`,
and `RubricConfig` as attributes of `judge.py`; those seams now live in
`judge/run.py`. Re-pointed, with no behavior assertion weakened.

The `RubricConfig.load_bundle` assertions became `from_paths` and are now
**stronger**: instead of confirming a manifest path was forwarded, they confirm
the three rubric files were resolved out of it
(`data/SI/rubric_manifest.json` → `data/SI/rubric.tsv` plus the two prompt
files). `test_main_loads_distinct_rubric_bundles_end_to_end` keeps its
real-parsing design and still proves `--rubrics` selects the bundle rather than
being a no-op.

## Validation

- Full non-live suite: **1,023 passed**, 8 deselected — same count as the
  `generate` branch, so nothing regressed
- Coverage: **74.90%** (gate: 30%)
- Ruff format/check: no new findings (3 pre-existing `E501`s in
  `judge/response_models.py`, untouched)
- Pyright on `judge.py`, `judge/run.py`, `utils/rubric_manifest.py`: **0 errors**,
  verified against a stashed baseline


judge/rubric_config.py
```diff
@@ -85,11 +85,46 @@ async def load(
...
+                Path(rubric_folder) / rubric_prompt_beginning_file
+            ),
+            question_prompt_file=str(Path(rubric_folder) / question_prompt_file),
+            sep=sep,
         )
-        question_prompt_path = Path(rubric_folder) / question_prompt_file
+
+    @classmethod
+    async def from_paths(
+        cls,
```
luca-belli, line 99: why cls, then *?

judge/run.py
```diff
@@ -0,0 +1,125 @@
...
+        Tuple of (results, output_folder) where output_folder is where the
+        evaluations were written.
+
+    Raises:
+        ValueError: If the output target is not exactly one of `output_root`
+            or `output_folder`.
+        FileNotFoundError: If a rubric file or the transcripts directory is
+            missing.
+    """
+    if (output_root is None) == (output_folder is None):
```
luca-belli: should't this be resolved by the CLI?

tests/unit/judge/test_judge_cli.py
```diff
@@ -144,14 +145,18 @@ async def test_main_single_conversation_calls_judge_single(self):
...
             LLMJudge.return_value = "judge_instance"
 
             result = await main(args)
 
-            RubricConfig.load_bundle.assert_called_once_with(
-                "data/SI/rubric_manifest.json"
+            # main() resolves the manifest to concrete paths, then the rubric is
+            # built from those paths rather than from the manifest itself.
+            RubricConfig.from_paths.assert_called_once_with(
+                rubric_file="data/SI/rubric.tsv",
```
luca-belli: is there a reason why this and not a fixture?

utils/rubric_manifest.py
```diff
@@ -79,6 +79,30 @@ async def load_manifest_personas(manifest_path: str) -> list[str]:
     return [str(manifest_dir / p) for p in manifest.get("personas", [])]
 
 
+async def load_manifest_rubric_paths(manifest_path: str) -> dict[str, str]:
+    """Resolve a manifest's three rubric files relative to the manifest.
```
luca-belli, line 83: add a clear example, so one doesn't need to know what a manifest it

eg say `if this is the CLI, it does this etc`

judge.py
```diff
@@ -158,8 +159,62 @@ def get_parser() -> argparse.ArgumentParser:
     return parser
 
 
+def _resolve_output_target(
```
luca-belli: shouldn't this be in util?
and if it's cli policy, should it be resolved upstream?

judge/rubric_config.py
luca-belli: `cls` is positional because `@classmethod` binds it that way — Python passes the class as the first positional argument, so it can't sit behind the `*`. The `*` starts one argument later and makes everything a caller supplies keyword-only.

That part is deliberate: the three parameters are all `str` and all paths, so `from_paths(rubric, prompt, question)` would silently accept them in any order. Keyword-only means a swap is a `TypeError` at the call site rather than a confusing `FileNotFoundError` (or worse, a successful load of the wrong file into the wrong slot). `load()` above keeps positional parameters because it is the pre-existing signature with defaults and callers already pass it positionally.

judge/run.py
luca-belli: You're right, and it pointed at the signature rather than the logic. Fixed in 23c1dd97.

The CLI *did* decide — `_resolve_output_target` in `judge.py` — but the decision arrived as two `Optional[str]`s where exactly one had to be set, which is a shape that can still be wrong. So the domain re-validated it (this guard) and re-branched on it (the `output_target` dict below). And `vera judge` never resumes, so it would have passed `output_folder=None` at every single call site: `output_folder` exists purely to serve the legacy script's resume mode. That is exactly the leak the module docstring claims this function prevents.

Now:

```python
output_dir: str,
is_existing_run: bool,
```

One required path plus a flag saying which of the two things it is. `vera judge` passes `is_existing_run=False` and never thinks about the other mode; `judge.py` passes `True` only under `--resume`. The guard and its `ValueError` docstring entry are deleted — the illegal state is no longer representable. The dict below stays, but it is now translating a resolved flag into `judge_conversations`' existing either/or kwargs, which is genuinely this layer's job.

tests/unit/judge/test_judge_cli.py
luca-belli: Two things were tangled there, and the second one was the real problem. Fixed in 23c1dd97.

The literal duplication was the smaller issue. The bigger one: these tests asserted against `data/SI/rubric.tsv` et al — real production rubric data — so renaming a shipped rubric file would have broken a CLI plumbing test that has nothing to do with rubric content. They only touched `data/SI/` because they relied on the default `--rubrics` value.

They now pass `tests/fixtures/rubric_manifest_simple.json` explicitly and assert against a shared `FIXTURE_RUBRIC_PATHS` constant. That the default *is* `data/SI/rubric_manifest.json` is still asserted, once, in `test_defaults` where it belongs.

I used a module constant rather than a pytest fixture because the value is a fixed expected literal with no setup and no teardown — a fixture would add an indirection without buying anything. Happy to switch if you'd rather it be uniform with the rest of the suite.

utils/rubric_manifest.py
luca-belli: Added a worked example in 23c1dd97 — it now shows the input manifest and the returned dict side by side, so the manifest-relative rule reads without knowing what a manifest is:

```
A manifest names its files by bare filename, and those names mean "beside
this manifest" -- never relative to $ROOT or the working directory. So
given `judge.py --rubrics data/SI/rubric_manifest.json` containing::

    {"rubric_file": "rubric.tsv",
     "rubric_prompt_beginning_file": "rubric_prompt_beginning.txt",
     "question_prompt_file": "question_prompt.txt"}

this returns::

    {"rubric_file": "data/SI/rubric.tsv",
     "rubric_prompt_beginning_file": "data/SI/rubric_prompt_beginning.txt",
     "question_prompt_file": "data/SI/question_prompt.txt"}
```

judge.py
luca-belli: On "resolved upstream" — it is; `main()` calls this before touching the domain, and `run_judging` now receives the answer as a single unambiguous value rather than a pair it has to re-check (see the `judge/run.py` thread).

On "should this be in util" — I argued myself out of it, and the reasoning is now recorded in the docstring. Every branch in here is legacy:

- the `<gen_run>/evaluations` default and the flat-folder stderr note encode the `p_*`/`j_*` layout that **Phase 3 retires** (`architecture.md` line 451: `utils/naming.py` is rewritten for the `c_`/`u_`/`j_` scheme)
- the resume check encodes a `j_*__*` basename rule that **`vera resume` will not use** — it reads `config.json`/`state.json`/`.sha256`, and the same doc says old-layout runs cannot be resumed at all because those files never existed

So the one part that looks reusable is the part that specifically does not carry over. And `judge.py` is retained *solely* as the resume entry point until `vera resume` lands, at which point it is deleted (line 449, "no new work targets it"). Moving this to `utils/` would give the conventions `vera` is replacing a home that outlives them.

That is the same test that sent `load_manifest_rubric_paths` **into** `utils/` last round — the manifest-relative rule is a property of the format, so both CLIs must agree on it. Output location is a property of each CLI's own UX, so `vera judge` resolves its own. `vera_cli/generate.py` already does exactly that at line 287.

luca-belli:
## Review round 2 — addressed in 23c1dd97

Replied inline to all five. Three changed code, two are answers.

### The substantive one

Your `judge/run.py:79` and `judge.py:162` comments were really one question — *shouldn't the CLI make all the decisions?* — and the answer split them apart along a line worth writing down, since the whole point of this PR is where that boundary sits:

| | belongs to | because |
|---|---|---|
| **Policy** — defaults, output location, model shorthand, whether `--resume` is legal | the CLI that offers it | each CLI's own UX; `vera judge` should not inherit `judge.py`'s |
| **Format rules** — "manifest paths resolve beside the manifest", "a run folder is named `j_*__*`" | shared code | not a decision any CLI gets to make; both must agree |

That test cuts both ways, which is why the two comments resolve in opposite directions:

- **`load_manifest_rubric_paths` moved *into* `utils/`** last round — format rule, outlives `judge.py`. ✅ correct call.
- **`_resolve_output_target` stays in `judge.py`** — pure policy, and *legacy* policy at that. Phase 3 retires the `p_*`/`j_*` layout its defaults encode, and `vera resume` reads `config.json`/`state.json` rather than the `j_*__*` basename its resume check validates. The one part that looked reusable is the part that specifically doesn't carry over. Reasoning is now in the docstring so the next reader doesn't re-litigate it.

### What actually changed

**`run_judging`'s output parameter** — this was the real find. It took two `Optional[str]`s where exactly one had to be set, so the domain re-validated (the guard) and re-branched (the dict) on a choice `main()` had already made. Worse, `output_folder` exists *only* to serve resume, which `vera judge` never does — so `vera` would have passed `output_folder=None` at every call site. The legacy script's argument convention, sitting in the seam whose docstring promises to keep it out.

```diff
-    output_root: Optional[str],
-    output_folder: Optional[str],
+    output_dir: str,
+    is_existing_run: bool,
```

```diff
-    if (output_root is None) == (output_folder is None):
-        raise ValueError(...)
```

The illegal state is no longer representable, and the `ValueError` leaves the docstring with it.

**`load_manifest_rubric_paths` docstring** — worked example, input manifest beside returned dict, so the manifest-relative rule reads without prior knowledge.

**Judge CLI tests** — they asserted against `data/SI/rubric.tsv` and friends, so renaming a shipped rubric file would have broken a CLI *plumbing* test that has nothing to do with rubric content. They now pass a `tests/fixtures/` manifest explicitly. That the default *is* `data/SI/rubric_manifest.json` is still asserted once, in `test_defaults`.

### Answered, unchanged

`from_paths(cls, *, ...)` — `cls` can't sit behind the `*`; `@classmethod` binds it positionally. The keyword-only part is deliberate: three same-typed path strings would otherwise swap silently and load the wrong file into the wrong slot.

### Validation

Unchanged from the PR description — **1,023 passed**, 8 deselected (same count, so the signature change is behavior-preserving), coverage **74.91%**, Pyright **0 errors** on `judge.py` / `judge/run.py` / `utils/rubric_manifest.py`, ruff clean.

utils/rubric_manifest.py
```diff
@@ -79,6 +79,44 @@ async def load_manifest_personas(manifest_path: str) -> list[str]:
...
+    """
+    manifest = await load_manifest(manifest_path)
+    manifest_dir = Path(manifest_path).parent
+    return {
+        "rubric_file": str(manifest_dir / manifest["rubric_file"]),
+        "rubric_prompt_beginning_file": str(
+            manifest_dir / manifest["rubric_prompt_beginning_file"]
+        ),
+        "question_prompt_file": str(manifest_dir / manifest["question_prompt_file"]),
+    }
```
jgieringer, line 152: If the files don't exist or have a typo from the manifest file, we'll get a KeyError
This is more of a nit, but it'd be nice to .get these and if None then raise a specific error about what's missing :bowtie:

jgieringer (approved):
Just a small nit - LGTM!

utils/rubric_manifest.py
luca-belli: good point thank you!

luca-belli: Good catch, fixed in b68eed2a — though the diagnosis landed one step over from the `KeyError`.

The three *keys* can't `KeyError`: `load_manifest` rejects a missing key with a `ValueError` naming it (`rubric_manifest.py:54`), and every resolver delegates there. But your "or have a typo" half was exactly right — nothing checked the *values*:

- `""` resolved to the containing directory, so you got `Rubric file not found: data/SI` — a missing *folder*
- `null` raised a bare `TypeError` out of `Path.__truediv__`
- a typo'd filename named the resolved path, but not the manifest that named it or the field to go fix

The check now lives in the resolver rather than in `RubricConfig.from_paths`, because `from_paths` holds three bare strings with no memory of where they came from — it structurally can't name the manifest however you word its message. The resolver is the last place that provenance is in scope:

```
Rubric bundle manifest data/SI/rubric_manifest.json: field 'rubric_file'
names 'rubirc.tsv', which is not a file beside the manifest: data/SI/rubirc.tsv
```

`is_file` rather than `exists`, so a directory-valued field is rejected here too instead of surfacing later as a missing folder.

Kept it dependency-free: `utils/` is the leaf layer, so it mirrors the shape of `vera_cli/targets.py`'s `resolve_file` with plain `ValueError`/`FileNotFoundError` rather than importing `existing_file`/`ConfigError`, which would invert the layer model. (`ConfigError` subclasses `ValueError`, so they stay catch-compatible.)

Also added the first direct tests for `load_manifest_rubric_paths`, which had none — happy path plus each rejected value shape.

One thing I deliberately left out: `personas` has the same unvalidated-value gap on the `generate.py` path, and no existence check anywhere. Fixing it changes generation behavior, so it's tracked separately in #208.

PR #195, opened 2026-08-14
Feat: add unified vera judge command

luca-belli:
## Summary

Adds the second unified CLI feature: `vera judge`.

Stacked on #194, which exposed the resolved-value judging entry point this
calls. Design decisions it implements were settled in #193.

## Review guide

`vera_cli/judge.py` is the same spine as `vera_cli/generate.py`, in the same
order: `run` → `resolve_configs` → `_from_cli`/`_from_config` → `_execute` →
`register`. Nothing after `resolve_configs` reads an `argparse.Namespace`;
nothing before `_execute` touches the judging domain. `register` sits last
because it is a declaration rather than a step — ~95 lines of flag definitions
should not sit between the reader and the logic.

```bash
uv run python vera.py judge -j gpt-4o:2 --conversations output/p_*__run --target SI
uv run python vera.py judge -j gpt-4o:2 --conversations output/p_*__run --rubric SI
```

Both resolve to byte-identical configs, exactly as `--target`/`--personas` do for
`generate`.

**`resolve_input` was reused unchanged.** The per-command parameterization of the
config-or-flags rule needed no modification to serve a second command, and
`ResolvedTarget`'s rubric fields — validated since #191, unused until now —
needed no new manifest code.

## Three things that were not a mirror

**`RunConfig` is multi-section.** `generation` and `judging` are both optional,
at least one required. `to_dict` **omits** absent sections rather than emitting
null: its output doubles as input config, and each command rejects top-level keys
it does not own, so a null section would make `generate --print` emit something
`generate` itself refuses. Covered by two tests, including a round-trip through
`VERA_RUN_CONFIG`.

**`generation_persona_sets` → `targets_from_config`.** It now owns only the two
rules both commands share — the `target: "all"` fan-out and target/explicit-field
mutual exclusion. Projecting a resolved target onto each command's fields stays
in the command module, avoiding two near-copies of the exclusivity logic;
`generate` calls its own one-line `_persona_sets`, `judge` its own rubric branch.

**Provider parameters return to the CLI.** `--judge-params` restores what
CHANGELOG v1.2 documents as `-jep` in its published migration instructions.
Parameters are supplied per role, as the legacy scripts accepted, and stored per
model, so `--print` shows what each model will actually use.

The matching `--user-params`/`--chatbot-params` for `generate` were originally
here too, closing the same gap on the generation side. **They have moved to
`feat/generate-model-params`**, stacked on this branch, because a generate-side
user-facing feature does not belong in a judging PR. They land there with the
three tests they were missing. Note this is a real capability gap rather than a
new feature: legacy `generate.py` has `-uep`/`-pep`, and Phase 1 deletes it, so
the branch needs to merge before that happens.

## Judge-specific behavior (per #193)

- **`--target all` rejected** until Phase 4 adds `evaluations/<target>/`; until
  then N rubric runs share one output folder, distinguishable only by timestamp.
- **`--conversations` takes exactly one folder** — judge separately, combine with
  `vera pool`. List-shaped per AD-20.
- **No single-conversation mode.** Judge a folder containing one conversation.
- **No `--resume` yet, and no exemption for `judge.py`.** #193 had carved
  `judge.py` out of Phase 1's deletion list to serve as a resume entry point;
  that carve-out is removed here. Porting the legacy per-stage skip to *both*
  commands is now stated as Phase 1 work instead, since both domains already
  accept `resume` and only CLI surface is missing. It is a different feature
  from the deferred, `state.json`-backed `vera resume`.
- **`--sample` is the shared debug cap** — conversations for judge, personas per
  file for generate — so `InvocationConfig` stays uniform rather than growing a
  per-command cap.
- **No `-c`.** Judging is decoupled from chatbot selection, and `-c` meant
  `--conversation` in legacy `judge.py`.

## Two ceilings found in the domain

`JudgingConfig` rejects inputs `run_judging` cannot honor, rather than accepting
them and silently doing something else:

- **Repeated judge model names** — the domain keys models by name, so duplicates
  would collapse.
- **Differing per-model provider parameters** — `run_judging` takes one parameter
  dict for the whole run. The schema is already shaped for per-model params, so
  this check can be deleted once the domain supports them.

## Breaking change

`vera judge` drops legacy `judge.py`'s fallback of writing evaluations to
`evaluations/` **relative to the working directory** when the input is a flat
transcript folder rather than a generation run. It errors and requires
`-o/--output`.

The default for a generation run is unchanged — `<conversation run>/evaluations/`,
beside the transcripts. Old flat folders stay readable with an explicit `-o`,
satisfying the read-old-data guarantee in `architecture.md`. Dropped because the
fallback detached evaluations from the conversations that produced them, and
because a bare relative path resolves against the working directory on the CLI
but against the repository root in a config.

Recorded in `CHANGELOG.md` under `[Unreleased] → Breaking / migration` and as an
`Acknowledged compatibility break` in architecture.md's Phase 1 row.

## Review round

Ten inline comments, all replied to; every one implying a change has landed.

**Placement and duplication**
- `_value` was byte-identical in both command modules, differing only in which
  `DEFAULTS` global it closed over implicitly. Now `config.flag_value`, taking
  `defaults` explicitly. Call sites are more verbose deliberately — a per-module
  one-line wrapper would put the same coupling back, just smaller.
- **`targets.py` had grown a second concern.** Its docstring says "discovery and
  validation of complete target manifests", but it also held `config_path`,
  `config_paths`, `config_dir`, and `rubrics_from_config` — which apply one rule
  (a config path is repo-relative and must exist) to whatever field a command
  hands them, and are thin wrappers over helpers already in `config.py`. All
  four moved there. `targets.py` is now only about turning a target *name* into
  a bundle.
- `judge.py` reordered to match `generate.py`, verified as a pure move by
  diffing sorted file content before and after.
- `_run_config` in both commands, and `targets_from_config`, are fully
  keyword-required; `invocation` was positional for no reason.
- `generate`'s persona-sets call site is one call again. Generalizing the target
  rules for `judge` was right, but it had been done by inlining a 28-line branch
  into `generate.py`, which made that module longer to serve a different command.

**`RubricFiles` is now resolved by construction.** Its docstring defines it as
"the resolved form", but the config path built it **twice** per entry — once from
raw repo-relative strings via `from_dict`, then again from those, resolved. The
type was transiently inhabited in the state its own name rules out, with the
first instance serving as a parsing scratchpad. Resolution now happens before
construction, and `from_dict` is deleted: it was the only route to the unresolved
state and had no callers left. Two tests pin the guarantee, including one that
checks *all three* paths are absolute — the existing test asserted only
`rubric_file`, so it would have passed with both prompt paths unresolved.

**`RunConfig` now explains what `invocation` is.** `generation`/`judging` say what
run to perform and define its identity; `invocation` says how one execution
behaved. Both persist into `config.json`; only the first is hashed.

## Merged `feat/VERA_2.0`

Brought onto this branch via #194 so review happens against the real base.

One textual conflict, resolved by keeping this branch's `generation is None`
guard while taking the incoming `register`-last placement.

**One semantic conflict git could not see, because it spanned files.** #194
replaced `run_judging`'s `output_root`/`output_folder` pair with an explicit
`output_dir` + `is_existing_run`, and `vera_cli/judge.py` still called the old
signature. Nothing conflicted textually, so the merge reported success and the
code was broken; pyright and the suite both caught it. `vera judge` always mints
a new `j_*` run, so `is_existing_run` is always False.

## Validation

- Full non-live suite: **1,060 passed**, 8 deselected
- Coverage: **75.49%** (gate: 30%)
- Ruff: no new findings (5 pre-existing `E501`s in `distribute_files.py` and
  `judge/response_models.py`, present on the base branch too)
- Pyright on `vera.py`, `vera_cli/`, `utils/config_schema.py`, `judge/run.py`,
  `judge.py`: **0 errors**
- Manually confirmed: `--target`/`--rubric` parity, `--target all` rejected, flat
  folder errors without `-o` and works with it, judge config round-trips through
  `VERA_RUN_CONFIG` byte-identically, and **`generate`'s round-trip is still
  byte-identical with no `judging` key** — the regression risk the multi-section
  schema introduced


utils/config_schema.py
```diff
@@ -192,19 +228,117 @@ def to_dict(self) -> dict[str, Any]:
...
+        names = [model.name for model in self.models]
+        if len(set(names)) != len(names):
+            raise ValueError(
+                "judging.models must not repeat a model name; use repeats to run "
+                "several instances of the same model"
+            )
+        # The judging domain takes one provider-parameter dict for the whole run,
+        # so per-model parameters cannot be honored yet. Reject them rather than
+        # silently applying one model's parameters to all. Drop this check when
+        # the domain accepts per-model parameters.
```
luca-belli, line 282: @jgieringer do we think that's fine? Or do we do combined judging runs with different params?

utils/config_schema.py
```diff
@@ -192,19 +228,117 @@ def to_dict(self) -> dict[str, Any]:
...
+
+    Holds one section per command taking part in the run. `generate` populates
+    `generation`, `judge` populates `judging`, and a later `pipeline` populates
+    both; at least one is required.
 
     `--target all` resolves to one `RunConfig` per target; every other input
     resolves to exactly one.
     """
 
     invocation: InvocationConfig
```
luca-belli, line 353: what's an invocation , and how is that different from Generation+Judging config?

vera_cli/generate.py
```diff
@@ -170,6 +175,20 @@ def register(subparsers: argparse._SubParsersAction) -> None:
             "(default: one session, using the chatbot's own session type)"
         ),
     )
+    parser.add_argument(
```
luca-belli: why are we adding generation code on a judging PR?

luca-belli: same question for the rest of the file

vera_cli/judge.py
```diff
@@ -0,0 +1,498 @@
...
+
+1. `register` declares the flags and attaches `run` as the subparser's handler.
+2. `run` is the entry point `vera.py` dispatches to.
+3. `resolve_configs` picks one input form and produces canonical `RunConfig`s.
+4. `_execute` hands each resolved run to the judging domain.
+
+Nothing below step 3 reads an `argparse.Namespace`, and nothing above it touches
+the judging domain.
+
+Two deliberate differences from legacy `judge.py`, both recorded in
```
luca-belli: remove this, future us won't care about this

vera_cli/judge.py
```diff
@@ -0,0 +1,498 @@
...
+            "generation run folder. Pass -o/--output to say where evaluations "
+            "should go."
+        )
+    return str((Path(generation_run) / "evaluations").resolve())
+
+
+def _from_cli(
+    args: argparse.Namespace, invocation: InvocationConfig
+) -> list[RunConfig]:
+    """Resolve CLI flags into one canonical run, applying CLI defaults.
```
luca-belli, line 131: remind me why dthose are not resoled by vera_cli/config?

vera_cli/judge.py
```diff
@@ -0,0 +1,498 @@
...
+                        "rubric_prompt_beginning_file",
+                        "question_prompt_file",
+                    )
+                }
+            )
+        )
+    return rubrics
+
+
+def _value(args: argparse.Namespace, field: str) -> Any:
```
luca-belli: that feels more general than just judge? should it be here?

vera_cli/judge.py
```diff
@@ -0,0 +1,498 @@
...
+    The path-list helpers in `targets` verify *files*; a conversations folder is
+    a directory, so it gets its own check.
+    """
+    resolved = Path(path_from_root(value))
+    if not resolved.is_dir():
+        raise ConfigError(f"{field} does not exist or is not a directory: {resolved}")
+    return str(resolved)
+
+
+def _rubrics_from_config(value: Any) -> list[RubricFiles]:
```
luca-belli: shouldn't that be resolved above?

vera_cli/judge.py
```diff
@@ -0,0 +1,498 @@
...
+
+
+def _value(args: argparse.Namespace, field: str) -> Any:
+    """Read a run-defining flag, falling back to its CLI default."""
+    return getattr(args, field, DEFAULTS[field])
+
+
+def _run_config(
+    invocation: InvocationConfig,
+    *,
```
luca-belli, line 307: why this here? what's special about invocation? we could make this a full keyword-requried function

vera_cli/targets.py
```diff
@@ -148,61 +148,62 @@ def resolve_file(field: str, path: object) -> str:
...
 
-def generation_persona_sets(
-    config: dict[str, object], generation: dict[str, object]
-) -> list[tuple[list[str], str]]:
-    """Resolve a config's persona inputs into one `(persona files, context)` pair
-    per run.
+def targets_from_config(
+    config: dict[str, object],
+    section: dict[str, object],
+    *,
```
luca-belli: why this is in the middle again?

utils/config_schema.py
luca-belli: Factual half is settled by the domain signature: \`judge/run.py:26\` takes \`judge_models: Dict[str, int]\` (name → instance count) and a **single** \`judge_model_extra_params: Dict[str, Any]\`.

So the domain physically cannot express per-model params today — this validation is not a policy choice, it is the signature. The check is doing the right thing by rejecting rather than silently applying one model's params to all.

The open question is whether the *domain* should change, which is a judging-domain decision rather than a CLI one. If combined runs with differing params are a real use case, that is a change to \`run_judging\`'s shape first, and this check is then deleted as its docstring already says.

utils/config_schema.py
luca-belli: The intended split: \`generation\`/\`judging\` describe **what run to perform** — reproducible, identity-defining, hashed into the run id. \`invocation\` describes **how this particular execution behaved** — \`debug\` and \`sample\`.

They are separate because they have different lifetimes: two runs with identical \`generation\` sections are the same run, even if one was invoked with \`--debug\`. \`ARCHITECTURE-SPINE.md:188\` covers this and calls out \`--sample\` as the one deliberate behavior-altering exception, which is why it is persisted rather than discarded.

If that is not obvious from the docstring in one line, the docstring is the bug. Will fix.

vera_cli/generate.py
luca-belli: Agreed in part. The +68 lines in this file are four distinct things and they do not all deserve the same verdict:

| Change | Verdict |
|---|---|
| \`--user-params\` / \`--chatbot-params\` (~25 lines) | **Scope creep.** A new user-facing \`generate\` feature. Judge needs nothing from it. Should be its own PR. |
| \`generation_persona_sets(...)\` → inlined \`targets_from_config\` + explicit branch (~35 lines) | **Right idea, wrong direction.** Generalizing the helper so judge can reuse the target/explicit-fields rule is correct, but it replaced a one-line call with an inline branch, so \`generate.py\` got *longer* in service of judge. The shared rule should stay behind one call. |
| \`run_config.generation is None\` guard | **Legitimate fallout.** \`RunConfig.generation\` had to become Optional so judging can populate its own section. Unavoidable. |
| \`ALLOWED_CONFIG_FIELDS\` comment | Legitimate, trivial. |

Plan: pull the params flags out into a separate PR, and restore generate's call site to a single call into the shared helper. The other two stay.

luca-belli: Covered in the reply above — broken down per change, with two of the four coming out.

vera_cli/judge.py
luca-belli: Pushing back on removal, agreeing on rewrite.

Future us will care, because **that sentence is about to become false.** `docs/architecture.md` Phase 1 deletes `judge.py`. The moment it does, "resuming stays available only through legacy `judge.py`" describes a capability that exists nowhere.

And this is not hypothetical: `judge/run.py:26` already takes `resume: bool`, so the judging domain has the capability fully wired — only the CLI surface is missing. Same shape on the generation side, where `vera generate` hardcodes `resume=False`. Deleting the legacy scripts without porting the flag is a capability regression on the most expensive operation in the system.

So: drop the "compared to legacy" framing, keep the fact. Something that still reads correctly after `judge.py` is gone — that the per-stage skip is not yet exposed, and that it is distinct from the deferred `vera resume` contract.

vera_cli/judge.py
luca-belli: Reading "those" as the three private helpers just above (`_reject_target_all`, `_rubric_from_target`, `_output_root`) — say if you meant something else.

No good reason for some of them. This is the same finding as your `config_dir` and `_rubrics_from_config` comments below, so treating it as one cleanup:

- `_rubrics_from_config` does per-field path resolution that `targets.py` already owns for every other path in the system → move.
- `config_dir` is the odd one out: its siblings `config_path`/`config_paths` live in `targets.py`, but this one sits in a command module, is **public** (no underscore), and has exactly one caller → move beside its siblings.
- `_reject_target_all` and `_output_root` are genuinely judge-specific (the second derives an evaluations path from a generation run folder) → stay.

So the rule: path resolution and the shared target rules move to the shared modules; judge-specific *policy* stays here.

vera_cli/judge.py
luca-belli: Confirmed — and stronger than "feels": `_value` is **byte-identical** in `vera_cli/generate.py` and `vera_cli/judge.py`.

```python
def _value(args: argparse.Namespace, field: str) -> Any:
    return getattr(args, field, DEFAULTS[field])
```

The only difference is that `DEFAULTS` is a module global each copy closes over implicitly — which is also why the duplication is easy to miss.

Moving it to `vera_cli/config.py` with `defaults` as an explicit parameter kills the copy and makes the dependency visible at the call site.

vera_cli/judge.py
luca-belli: Yes. `_rubrics_from_config` is doing per-field path resolution that `targets.py` already owns for every other path in the system, so it is re-implementing the shared concern locally.

Grouped with your `config_dir` / `_from_cli` comments as one cleanup — see the reply on the `_from_cli` thread for the split of what moves and what stays.

vera_cli/judge.py
luca-belli: Nothing is special about it — `_run_config(invocation, *, ...)` is positional-then-keyword by convention only, carried over from `generate.py`.

Agreed, make it fully keyword-required. Strictly better, costs nothing, and it removes the implication that `invocation` has some privileged status it does not have.

vera_cli/targets.py
luca-belli: Two separate things here, and the "again" is worth untangling.

**1. `targets_from_config` sitting between helpers in this file** — fair. It is the public entry point for the shared target rule and should sit with the other public functions (`target_catalog`, `resolve_target_manifest`, `target_manifest_paths`, `load_target`), not below `load_target`'s private helpers.

**2. The `register`-first ordering across the CLI modules** — this one is a *branch-age artifact*, not a decision anyone made in this PR.

`#191` reordered `vera_cli/generate.py` so `register` comes **last**, with the reasoning recorded in the module docstring: `register` is a declaration rather than a step, and ~100 lines of flag definitions should not sit between the reader and the logic.

`#194`/`#195` branched from `docs/judge-command-plan` *before* `#191` merged, so the `generate.py` on this branch is simply the pre-`#191` copy. Nothing here reverted anything.

The consequence still needs fixing though: after rebasing onto `feat/VERA_2.0`, `generate.py` will have `register` last while `judge.py` has it first — and `judge.py`'s docstring claims it is "Structured exactly like `vera_cli/generate.py`", which will then be false. Better to fix in this PR than to discover it post-merge.

vera_cli/judge.py
luca-belli: Correction to my reply above — I had the Phase 1 row wrong, and it flips the reasoning.

`feat/VERA_2.0` already carries #193's update, which says the opposite of what I claimed:

> delete `generate.py`/`run_pipeline.py` at the end of the phase. **`judge.py` is the one exception and outlives this phase:** `vera judge` ships without `--resume`, because the resume contract is deferred, so `judge.py` is retained *solely* as the resume entry point until `vera resume` exists.

So `judge.py` is **not** deleted in Phase 1, and the docstring sentence does not become false — it stays true by design.

My conclusion is unchanged but the reason is different, and better: this note is the in-code record of *why `judge.py` is deliberately being kept alive*. That is precisely what future-us needs when they find a legacy script that Phase 1 was supposed to have removed. Still worth rewriting rather than deleting — drop the "differences from legacy" framing, keep the statement that the per-stage skip is unexposed here and lives in `judge.py` until `vera resume` lands.

Separately, and worth raising on its own: the same row deletes `generate.py` with **no** equivalent exception, even though `vera generate` hardcodes `resume=False` while the generation domain already accepts `resume: bool`. So generation loses the capability at Phase 1 and judging does not. That asymmetry looks unintentional — either `generate` deserves the same exception, or both should get the per-stage flag ported.

luca-belli: Decision from @luca-belli: **no exception for judging.** `judge.py` is deleted in Phase 1 alongside `generate.py` and `run_pipeline.py`, like everything else.

That makes the original comment on this thread correct and my pushback moot — the "resuming stays available only through legacy `judge.py`" note describes a plan that is no longer the plan, so it comes out rather than getting rewritten.

The consequence is that both commands now need the capability, not just `generate`. Both domains already accept it — `judge/run.py:26` takes `resume: bool`, and `vera generate` hardcodes `resume=False` — so this is CLI surface only, no domain work.

Four places currently encode the exemption and all need to change:

| Location | Branch | Change |
|---|---|---|
| `docs/architecture.md` Phase 1 row | `feat/VERA_2.0` **and** re-edited here | Drop the `judge.py`-outlives-Phase-1 carve-out; restore it to the deletion list |
| `docs/vera-cli-use-cases.md:24-27` | `feat/VERA_2.0` **and** here | Drop the "`judge.py` is the one that lingers" paragraph |
| `vera_cli/README.md` judge row | here | Drop "no `--resume`" from the list of differences |
| `vera_cli/judge.py` module docstring | here | Remove the deferred-resume bullet |

Note the first two live on `feat/VERA_2.0` as well as on this branch, so they are not fully fixable from inside this PR — the base branch needs the same edit, or this PR carries it and the base inherits it on merge.

Replacing the carve-out: port the existing per-stage skip to both commands as a distinct flag rather than reusing the word `resume`, which stays reserved for the deferred `state.json`-backed subcommand (AD-18/23/24). Proposed shape, one flag instead of legacy's overloaded pair:

```
vera generate --into <run folder>     # continue this run, skip what is already done
vera judge    --into <run folder>
```

`--into` selects the run *and* implies the skip, and is mutually exclusive with `-o`. Legacy needs `-o <run folder> --resume` because the boolean says "skip" while an overloaded `-o` says "which run" — that overload is exactly what the resolution rule exists to prevent.

vera_cli/targets.py
luca-belli: Re-read the anchor and I think I answered the wrong question. The comment sits on the `*,` line, not on the `def` — and "again" lines up with your `_run_config` comment about `invocation` being positional. So I now read this as the same point: make it fully keyword-required.

Done in fa491c9e — `targets_from_config`, and `_run_config` in both command modules, are now `*`-first with nothing positional.

My earlier reply about file placement and the `register` ordering still stands as a separate observation, but it was not an answer to this. Say the word if you did mean placement and I will move the function too.

vera_cli/judge.py
luca-belli: Done in `ed4a53b0`, and your framing found something better than either option I had offered.

`RubricFiles`'s docstring already defines it as *"the resolved form"*. But this function was building it **twice** per entry:

```python
files = RubricFiles.from_dict(entry)      # unresolved — raw repo-relative config strings
rubrics.append(RubricFiles(**{            # resolved — rebuilt from the first
    field: config_path(getattr(files, field), ...) ...
}))
```

So the type was transiently inhabited in exactly the state its own name rules out, with the first instance serving as a parsing scratchpad. That was the real defect — not where the function lived.

Resolution now happens before construction:

- `validate_dict` checks an entry's shape without instantiating, and `field_names` derives the field list from the dataclass so a fourth rubric file cannot leave a hand-written check behind.
- **`from_dict` is deleted.** It was the only route to the unresolved state, and afterwards it had no callers. Keeping it for a hypothetical persisted-config reader would have preserved the hole for a caller that does not exist.
- The resolution moves to `vera_cli/targets.py` as `rubrics_from_config`, beside `config_path`/`config_paths`/`config_dir` — it applies the same rule those do. It cannot live with the dataclass in `utils/config_schema.py`, tidy as that would be, because resolution needs `path_from_root` and `utils/` is the leaf layer.

`vera judge` now only decides *which* rubric — target or explicit — and receives it resolved, which is what you were asking for. The two remaining construction sites both build from a `ResolvedTarget`, already verified by `load_target`.

Two tests pin the guarantee. One checks all three fields are absolute: the existing test asserted only `rubric_file`, so it would have passed with both prompt paths left unresolved — precisely what the two-step construction allowed.

vera_cli/targets.py
luca-belli: Both readings addressed, in `9e631875`.

**Keyword-only:** done earlier — `targets_from_config` and both `_run_config`s are now `*`-first.

**Placement:** you were right, and digging in found more than the one function. `targets.py`'s docstring says *"discovery and validation of complete target manifests"*, but the bottom of the file held `config_path`, `config_paths`, `config_dir`, and `rubrics_from_config` — none of which are about target manifests. They apply one rule (a config path is repo-relative and must exist) to whatever field a command hands them, and they are thin wrappers over `path_from_root`/`existing_file`, which already live in `config.py`.

So the module had two concerns, and `targets_from_config` sat exactly on the seam between them. It *was* in the middle. Moved all four to `config.py`; `targets.py` is now only about turning a target name into a bundle.

Worth owning: two of those four were mine, added earlier in this review round. I moved them toward the right *rule* but the wrong *module*, which made the split worse before it got better.

Dependency direction is unchanged — `targets.py` already imported `config.py` and never the reverse — so no cycles. Both module docstrings and the `vera_cli/README.md` layout table now state the split explicitly.

Also in the same commit: `judge.py` reordered to match `generate.py` (`register` last), so its "structured exactly like generate.py" docstring is true again. Verified as a pure move by diffing sorted file content before and after — the only content changes are the docstring and one comment.

utils/config_schema.py
jgieringer: @luca-belli Yeah this breaks for instances like gpt-5.4 with reasoning_effort=low and claude-sonnet-5 without reasoning.
Since the recommended run is just one judge, we're okay, but should be looked into for full judge customization!
wonder if it's something like remove --judge-extra-model-params and have them tagged onto each judge?
`claude-sonnet-5:2 gpt-5.4:1:reasoning_effort=low,verbosity=high`
or do something like index?
`-jep 1:reasoning_effort=low,verbosity=high`

utils/config_schema.py
```diff
@@ -78,6 +78,59 @@ def to_dict(self) -> dict[str, Any]:
...
+            if not isinstance(value, str) or not value:
+                raise ValueError(f"judging.rubrics {field.name} must be a path")
+
+    @classmethod
+    def field_names(cls) -> tuple[str, ...]:
+        """The three field names, in declaration order."""
+        return tuple(field.name for field in dataclasses.fields(cls))
+
+    @classmethod
+    def validate_dict(cls, data: dict[str, Any]) -> None:
```
jgieringer, line 137: should we do a validate_rubric that checks if the files exist?

jgieringer (approved):
Nice!

utils/config_schema.py
luca-belli: Pushing back on a `validate_rubric`, but the question found three real things — landed in `8984cb3b` and `fa8899be`.

**Why not a separate validator:** #194 settled this one layer down. `b68eed2a` asked the same question of the legacy manifest path and put the check in the *resolver*, explicitly not in `RubricConfig.from_paths`, because that type

> holds three bare strings with no memory of their origin

`RubricFiles` is the same shape. It can't name the manifest or config field a bad path came from, so its error would be strictly *worse* than what the boundary already emits:

```
judging.rubrics.rubric_file does not exist or is not a file: /…/data/SI/rubirc.tsv
question_prompt_file does not exist or is not a file: /…/target/question_prompt.txt
```

And existence is already checked exactly once per input form, so a second check would `stat()` the target route's files twice and report one condition two ways:

| Route | built at | existence owner |
|---|---|---|
| config `judging.rubrics` | `config.py:198` | `config_path` → `existing_file` |
| config `target:` | `judge.py:211` | `load_target` → `resolve_file` → `existing_file` |
| `--target` / `--rubric` | `judge.py:263` | same — `--rubric` is a target name, not a file path |

**What you did find.** The docstring claimed a guarantee the class didn't make. "Resolved" is two things, and only one was enforced:

- ***exists*** — the boundary's, and stays there. Now stated rather than implied.
- ***absolute*** — enforced by *nothing*. `__post_init__` checked non-empty `str`, so `"data/SI/rubric.tsv"` was accepted and later opened relative to the **working directory** — a different file from the one a config naming that same string means. That's the exact ambiguity this PR's breaking-change note is about, and it's the bug the old two-step construction had. Now checked, which is what the module docstring puts in scope (*"relative paths are absolute"*) while keeping files out of it (*"does not read files"*). Pure string inspection — no I/O, no `vera_cli` import.

`ValueError` not `FileNotFoundError`, per the split `b68eed2a` set: an unresolved path is a bad value *shape*, not a path that names nothing.

The docstring was also stale twice over — it named `vera_cli.targets.rubrics_from_config`, which moved to `vera_cli.config` in `9e631875`, and called it *"the one place"* construction happens when there were **three**. Two of those were the same three-field rename written twice, now one `_rubric_files` helper (`fa8899be`).

**Three tests, each confirmed to fail with its check removed.** The two existence ones are the first tests *anywhere* of `existing_file`'s failure branch on the `vera` path — `b68eed2a` had just added the equivalents for the legacy resolver, so the gap was newly visible. They assert the message, not just the exit code: every input error here exits 2, so a code-only assertion would keep passing if the existence check were deleted and something else rejected the config instead.

Two things I deliberately didn't do here:

- `GenerationConfig`'s path fields (`personas`, `output`, `persona_context_template`) have the identical gap. That's generation-side work in a judging PR, so it's #209 + a `TODO` line, following how `a7311a32` handled the sibling `personas` gap.
- Worth flagging since it's the deeper version of your concern: **none of this catches a rubric edited *after* resolution.** That's `TODO` line 1 — *"Config files should also have the sha of the files they use (e.g., rubrics) to avoid silent failures"* — and it's the one that would actually close the silent-failure hole.

Also retargeted this PR onto `feat/VERA_2.0` now that #194 is merged, and merged it in (`ab864330`, no conflicts).

vera_cli/judge.py
```diff
@@ -207,13 +208,7 @@ def _from_config(
...
                 "different rubrics would share one output folder"
             )
-        rubrics = [
-            RubricFiles(
-                rubric_file=targets[0].rubric,
-                rubric_prompt_beginning_file=targets[0].rubric_prompt_beginning,
-                question_prompt_file=targets[0].question_prompt,
-            )
-        ]
+        rubrics = [_rubric_files(targets[0])]
```
luca-belli, line 217: is this because only one rubric at time is allowed?

luca-belli: No — and the fact that it reads that way is a fair hit, so I've made the line say so in `8e43a77b`.

**Two different single-ness constraints stack here, three lines apart.** The guard above carries its reason in its error message; the list-of-one carried nothing, so it looked like a consequence of the guard.

| | why it's 1 | lifted by |
|---|---|---|
| `len(targets) != 1` (206) | N rubrics' evaluations would share one output folder, distinguishable only by timestamp | Phase 4's `evaluations/<target>/` segment |
| `rubrics = [...]` (211) | `JudgingConfig.__post_init__` accepts exactly one; the *list* is AD-20, list-shaped from day one so multi-rubric isn't a schema break | Phase 4's multi-rubric support |

They're genuinely independent, and the giveaway is that **they lift into different shapes.** AD-21:

> `--target all` produces **one canonical invocation per discovered target** and never merges different targets' personas, prompts, or rubrics.

So lifting the target guard doesn't make this a multi-rubric run — it makes it N runs, exactly as `generate` already does via `_persona_sets`. Line 211 survives that lift unchanged, just relocated:

```python
return [
    _run_config(..., rubrics=[_rubric_files(target)], ...)
    for target in targets
]
```

Multi-rubric is the other axis: *one* run, N rubrics, `rubrics` finally longer than 1. A config could eventually hit both at once — N targets × N rubrics each — which is why neither constraint can stand in for the other.

So to answer directly: the `[...]` wrapper is the multi-rubric restriction (AD-20 shape, Phase 4 lift), and it would still be `[...]` even after `--target all` starts working.

PR #196, opened 2026-08-14
Support custom base URL for major providers

jgieringer:
### Summary
- Adds ANTHROPIC_BASE_URL, OPENAI_BASE_URL, and GOOGLE_BASE_URL config support so all three chat providers can be routed through a custom endpoint (e.g. an internal LLM gateway), each forwarded to its LangChain client's base_url param when set.
- Fixes Gemini specifically: langchain-google-genai 3.1.0's underlying GAPIC client defaults to gRPC, which requires a bare hostname:port and silently hangs against a path-based REST gateway URL; forcing transport="rest" as a workaround then hit a second bug where that version's async client has no real async REST transport ('GenerateContentResponse' object can't be awaited). Resolved by upgrading langchain-google-genai to 4.3.4, which moves to Google's REST/async-native google-genai SDK — no transport workaround needed.
- Bumps langchain-anthropic (1.4.7 → 1.5.6) and langchain-openai (1.0.3 → 1.5.1, pulling openai SDK 2.8.1 → 3.0.0) for general dependency hygiene while in here — these two already worked correctly with a custom base URL since both providers are REST-only.

### Test plan
- [X] uv run pytest -m "not live" — 981 passed, no regressions
- [X] Live conversation generation through the gateway for Gemini (gemini-3.5-flash), Claude (claude-sonnet-5), and OpenAI (gpt-5.2) via generate.py — all three completed with real, coherent transcript output
- [X] New unit tests covering *_BASE_URL config loading and forwarding to each provider's LangChain client (test_config.py, test_claude_llm.py, test_gemini_llm.py)

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,8 +1,12 @@
 ANTHROPIC_API_KEY=your_anthropic_api_key_here
+# Optional custom Anthropic-compatible endpoint
+ANTHROPIC_BASE_URL=
 
 OPENAI_API_KEY=your_openai_api_key_here
 
 GOOGLE_API_KEY=your_google_api_key_here
+# Optional custom Gemini-compatible endpoint
+GOOGLE_BASE_URL=
 
 AZURE_API_KEY=your_azure_api_key_here
 AZURE_ENDPOINT=your_azure_endpoint_here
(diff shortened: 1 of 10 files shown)


luca-belli:
closing in favour of #197

PR #197, opened 2026-08-21
feat: support optional base URL for Claude, OpenAI, and Gemini clients

luca-belli:
## Summary

The Claude, OpenAI, and Gemini clients always called the providers' public APIs, so there was no way to route them through a proxy or gateway (e.g. LiteLLM). Setting a gateway key as `ANTHROPIC_API_KEY` sent that key to `api.anthropic.com`, which returns **HTTP 401**.

Adds `Config.get_base_url(provider)` (`llm_clients/config.py`), which reads a shared `API_BASE_URL` with per-provider `ANTHROPIC_BASE_URL` / `OPENAI_BASE_URL` / `GOOGLE_BASE_URL` overrides. Trailing slashes are stripped; an unknown provider name raises.

Each of the three clients passes `base_url` **only** when one is configured, so provider defaults are unchanged when it is not set — this is inert for anyone not using a gateway.

Verified the LangChain wrappers accept the parameter (`ChatAnthropic.anthropic_api_url` and `ChatOpenAI.openai_api_base` both alias `base_url`; `ChatGoogleGenerativeAI` has a native `base_url` field), and confirmed against real client objects rather than only mocks — all three end up with the URL set. Also confirmed `_filter_supported_params` does not strip it.

`.env.example` and the README setup section document the new vars, including the caveat that the matching `*_API_KEY` must then be a key the gateway accepts.

## Architecture checklist

- [x] Read [architecture.md](docs/architecture.md)
- [ ] OpenSpec change linked (required for multi-file or cross-boundary work)
- [x] `uv run pyright` passes
- [x] `uv run pytest -m "not live"` passes
- [ ] Label **`architecture-change`** if boundaries, entry points, or dependencies changed

Notes on the two unchecked boxes and one caveat on pyright:

- No OpenSpec change linked — this adds an optional config knob rather than crossing a boundary. Happy to add one if reviewers disagree.
- No `architecture-change` label: no boundaries, entry points, or dependencies changed.
- `uv run pyright` reports **0 errors in the files this PR touches**. Project-wide it reports 412, all pre-existing and concentrated in stray local dirs (`docs/README_to2/`, `tmp_tests/`) plus existing `llm_clients`/`judge`/`tests` errors that this branch does not introduce.

## Test plan

- [x] Local checks run
- [ ] CI green

8 new tests: 5 covering `get_base_url` (unset, shared fallback, per-provider override, trailing-slash stripping, unknown-provider error) and a pass/omit pair per client asserting `base_url` reaches the wrapper when configured and is absent when not.

`uv run pytest -m "not live"` — **972 passed**, coverage 75%.


jgieringer (dismissed):
I think this is fine, but definitely will need to update gemini in pyproject as older versions do not effectively support base_url
See #196 pyproject https://github.com/SpringCare/VERA-MH/pull/196/changes#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711

jgieringer (changes requested):
Update pyproject & uv.lock like #196 
https://github.com/SpringCare/VERA-MH/pull/196/changes#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711

luca-belli:
> Update pyproject & uv.lock like #196 https://github.com/SpringCare/VERA-MH/pull/196/changes#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711

Done!

jgieringer (approved):
LGTM!

PR #200, opened 2026-08-26
Feat: restore --user-params/--chatbot-params on vera generate

luca-belli:
## Summary

Restores CLI provider parameters for the generation-side models: `--user-params`
and `--chatbot-params` on `vera generate`.

Stacked on #195, which introduces the shared `models_from_cli` helper this uses
and the matching `--judge-params`.

## Why this is parity work, not a new feature

Legacy `generate.py` accepts `-uep`/`--user-agent-extra-params` and
`-pep`/`--provider-agent-extra-params`. `vera generate` as shipped in #191 has no
CLI equivalent — `ModelSpec.from_shorthand` always produces empty
`extra_params`, so `--config` is the only route to setting a temperature.

That is a capability gap of the same shape as the missing `--resume` skip, and it
matters on the same deadline: `docs/architecture.md` Phase 1 deletes
`generate.py`. **This needs to merge before that happens**, or the capability is
lost rather than migrated.

## Design

Parameters are supplied per **role** on the command line — one flag covering
every model of that role — which is what the legacy scripts accepted:

```bash
uv run python vera.py generate -c sonnet -u gpt:2 claude:1 --target SI \
  --user-params temperature=0.7,max_tokens=1000 \
  --chatbot-params temperature=0.1
```

The resolved form stays **per model**: each `ModelSpec` gets its own copy, so
`--print` shows exactly what each model will use, and a printed config can then
be edited one model at a time. Per-model differentiation on the command line is
deliberately not offered — the `name[:repeats]` shorthand has no room for it, and
config input already covers that case.

## Provenance

These two flags were originally written as part of #195. They were pulled out
because a generate-side user-facing feature does not belong in a judging PR, and
because they were the edits sitting *inside* the `register` block that #191 moved
to the end of the module — removing them first turned that branch's incoming
merge conflict from ~110 lines into a 5-line adjacency.

They arrive here with three tests they did not originally have.

## Tests

| Test | Asserts |
|---|---|
| `test_role_params_apply_to_every_model_of_that_role` | Role params reach every model of their role and no other, each model gets a **distinct** dict (so per-model editing of a printed config actually works), and the `name[:repeats]` shorthand is undisturbed |
| `test_role_params_default_to_empty_when_flags_absent` | No flags → `extra_params == {}` for both roles |
| `test_role_params_are_run_defining_so_config_rejects_them` | `--config` rejects them. This holds *automatically*: they are not in `INVOCATION_ONLY_FLAGS` and the rule is derived by subtraction, so no registration was needed — the test is what keeps that true if the classification ever changes |

## Validation

- Full non-live suite: **1,063 passed**, 8 deselected
- Coverage: **75.51%** (gate: 30%)
- Ruff and pyright: clean


vera_cli/generate.py
```diff
@@ -177,8 +179,10 @@ def _from_cli(
     return [
         _run_config(
             invocation=invocation,
-            chatbot=ModelSpec.from_shorthand(chatbot),
-            users=[ModelSpec.from_shorthand(user) for user in users],
+            chatbot=models_from_cli([chatbot], getattr(args, "chatbot_params", None))[
+                0
```
luca-belli: why are we limiting to the first item here?

jgieringer (approved):
LGTM

PR #201, opened 2026-08-27
Feat: add --into to continue an existing run

luca-belli:
## Summary

Adds `--into <run folder>` to `vera generate` and `vera judge`: continue an
existing run, skipping work whose output is already on disk.

Stacked on #200. Closes the last capability gap before Phase 1 can delete the
legacy scripts.

## Why

`generate.py --resume` and `judge.py --resume` both work today. Neither `vera`
command exposed the capability, so deleting the scripts — which
`docs/architecture.md` Phase 1 now says happens with no exceptions — would have
lost it on the two most expensive operations in the system: a run that dies 80%
through could not be salvaged until Phase 3.

**Both domains already accept `resume`.** `generate` was passing a hardcoded
`resume=False`, and `judge` the same. So this is CLI surface only — no domain
code changed.

## Three decisions

**One flag, not a pair.** Legacy needs `-o <run folder> --resume`, because the
boolean says "skip" while an overloaded `-o` says "which run". `--into` says
both, and is mutually exclusive with `-o`, so `-o` keeps meaning exactly one
thing: a parent to mint a new run under.

```bash
vera generate -c sonnet -u gpt:1 --target SI --into output/p_gpt__a_sonnet__t30__r1__<ts>
vera judge    -j gpt-4o --conversations <run>/conversations --target SI --into <run>/evaluations/j_<...>
```

**Named `--into`, not `--resume`.** That word stays reserved for the deferred,
`state.json`-backed `vera resume` (AD-18/23/24), which is a different feature.
This one is stateless: it re-derives progress from the files already written,
which is why it needs none of Phase 3's artifacts and can ship now.

**Invocation-only, not run-defining.** `--into` says nothing about *which* run
this is — a run completed in one go and the same run finished across two
invocations are the same run, which is the whole point of being able to continue
one. Two consequences, and they are why the classification matters:

- **No existing config breaks.** `_from_config` requires every run-defining
  behavior field, so had this landed in `GenerationConfig`/`JudgingConfig`, every
  stored `--config` document would have stopped resolving until it added the
  field. As an `InvocationConfig` field defaulting to `None`, the change is
  purely additive.
- A config naming a folder would otherwise be **usable exactly once** — the
  second time, the folder is already complete.

AD-17 says a new behavior-altering flag needs its own named exception rather
than an implicit ride on `--sample`'s, so `ARCHITECTURE-SPINE.md` and
`vera-cli-use-cases.md` now name both.

## Tests

| Test | Asserts |
|---|---|
| `test_into_continues_an_existing_run_folder` | The domain receives `resume=True` and the folder as `output_folder` |
| `test_into_and_output_are_mutually_exclusive` | `-o` + `--into` is rejected by the parser |
| `test_into_rejects_a_folder_that_does_not_exist` | Continuing a run never started is a typo, not a request to create one |
| `test_into_is_invocation_only_so_it_may_accompany_config` | `--config` accepts it, where any run-defining flag would be refused |
| `test_configs_without_into_still_resolve` | The additive-not-breaking guarantee |
| `test_into_continues_an_existing_evaluation_run` | Judge maps it onto `output_dir` + `is_existing_run` + `resume` together |

## Known coupling, accepted

The run-folder validation in `generate_conversations/main.py` parses the legacy
`p_*`/`a_*` naming scheme, so `--into` inherits that and needs updating when
Phase 3 rewrites naming to `c_*`/`u_*`. That coupling already exists in
`--resume`; this does not add it, it keeps it alive slightly longer.

At Phase 3 the flag will also need AD-24's collision-check exemption, since it
writes into an existing run root.

## Validation

- Full non-live suite: **1,069 passed**, 8 deselected
- Coverage: **75.55%** (gate: 30%)
- Ruff and pyright clean on every touched file
- Manually confirmed: `--into` round-trips through `--print`, `-o` exclusion
  errors, a missing folder errors, and omitting the flag emits `"into": null`
  with behavior unchanged


docs/vera-cli-use-cases.md
```diff
@@ -207,13 +207,23 @@ vera pipeline --config run.json --sample 2
...
 just for smoke testing.
 
-**`--sample` is the sole behavior-altering exception to the CLI/`--config`
-either-or rule** (AD-17 in
-[ARCHITECTURE-SPINE.md](./ARCHITECTURE-SPINE.md)): it caps how much of the
-already-resolved lists get used rather than replacing those lists. The executed
-value is recorded in the run's persisted `config.json` invocation metadata, so
-`vera resume` retains the same sampled scope. `--debug` is recorded alongside
-it; `--print` creates no run and therefore has nothing to persist.
+**`--sample` and `--into` are the two behavior-altering exceptions to the
```
luca-belli: do we think it's ok to have two separate exceptions, or does it mean we have a weak design?

vera_cli/config.py
```diff
@@ -205,6 +205,24 @@ def rubrics_from_config(value: object) -> list[RubricFiles]:
     return rubrics
 
 
+def _resolved_into(value: Any, *, from_config: bool) -> str | None:
```
luca-belli: why Any? isn't that too generic?

also stop using `*` in the middle of everything, either at the beginning or nowhere

vera_cli/config.py
```diff
@@ -205,6 +205,24 @@ def rubrics_from_config(value: object) -> list[RubricFiles]:
...
+    """Resolve `--into`/`invocation.into` to an existing run folder, or `None`.
+
+    Follows the same split as every other path: a CLI value resolves against the
+    working directory, a config-stated one against the repository root. The
+    folder must already exist — continuing a run that was never started is a
+    typo, not a request to create one.
+    """
+    if not value:
```
luca-belli, line 242: add comments explaining what's happening in each case (ie why is `not value`? what case is that)

vera_cli/config.py
```diff
@@ -266,17 +284,21 @@ def resolve_input(
         value = config.get("invocation", {})
         if not isinstance(value, dict):
             raise ConfigError("invocation must be an object")
-        unknown = set(value).difference({"debug", "sample"})
+        unknown = set(value).difference({"debug", "into", "sample"})
```
luca-belli: should this list be maintained somewhere instead of being hidden in line 287 of a random file?

vera_cli/config.py
```diff
@@ -266,17 +284,21 @@ def resolve_input(
...
         if unknown:
             raise ConfigError(
                 f"unknown invocation field(s): {', '.join(sorted(unknown))}"
             )
         persisted = value
 
     try:
+        into = getattr(args, "into", persisted.get("into"))
```
luca-belli, line 333: why is this? add a comment

vera_cli/config.py
```diff
@@ -266,17 +284,21 @@ def resolve_input(
...
         invocation = InvocationConfig(
             debug=getattr(args, "debug", persisted.get("debug", False)),
             sample=getattr(args, "sample", persisted.get("sample")),
+            # A CLI path, so it resolves against the working directory like
+            # `-o` does; a config-stated one resolves against the repo root.
```
luca-belli: use full form instead of `-o`

vera_cli/generate.py
```diff
@@ -79,7 +79,7 @@
 # run-defining by subtraction (`resolve_input` derives it from the parsed
 # namespace), so a newly added flag is subject to the config-or-flags rule
 # automatically rather than needing to be listed somewhere second.
-INVOCATION_ONLY_FLAGS = frozenset({"config", "sample", "debug", "print_only"})
+INVOCATION_ONLY_FLAGS = frozenset({"config", "sample", "debug", "print_only", "into"})
```
luca-belli: this is what i was mentioning in the other comment as well

vera_cli/config.py
luca-belli: and in the rest as well

docs/vera-cli-use-cases.md
luca-belli: Pushed a reframing in 3447f5e8 — I think the two-exceptions phrasing was the weak part, not the design.

The either/or rule only ever governed information that says *which run this is*. An invocation-only control supplies none of that, so it was never an exception to the rule; it is the category the rule already provides for. `--debug` has always been in it and nobody called it an exception. `--sample` and `--into` only felt exceptional because they alter behavior rather than only observability — but they alter *this execution's* behavior, not the run's identity.

What did smell was that the category was enumerated by hand in three places. It is now exactly the fields of `InvocationConfig` (see the reply on the other thread), so membership is structural rather than asserted. AD-17 and vera-cli-use-cases.md say category, not exception, and keep the requirement that a behavior-altering control be named and justified before joining it — that part was worth keeping, since "is this run-defining?" is a judgment call worth making in review.

vera_cli/config.py
luca-belli: Both fixed in 3447f5e8.

`Any` → `object`: `Any` silences the type checker, `object` still forces narrowing, which is what this function does. It cannot claim `str` on the way in because the value arrives unvalidated from two places — argparse (always a str when present) and a parsed JSON document (any JSON type). Matches `rubrics_from_config` right above it, which already took `object`.

`*`: now at the front, so everything is keyword-only.

vera_cli/config.py
luca-belli: Done in 3447f5e8 — each branch says which case it covers: the three ways to mean "no folder" (flag omitted, `"into": null`, empty string) and why they are all silent rather than errors; that the non-string branch is only reachable from a config document, which is why the message names the config field; and that the `is_dir` check happens at resolution time because creating the folder would silently start a fresh run under a name the user expected to already hold results.

vera_cli/config.py
luca-belli: Agreed — it was written down three times: that set, plus an identical `INVOCATION_ONLY_FLAGS` in `generate.py` and another in `judge.py`.

3447f5e8 makes `InvocationConfig` the single definition, since that dataclass already *is* what invocation-only means. It grows a `field_names()` classmethod (same pattern `RubricFiles` already uses) and all three readers derive from it. `INVOCATION_ONLY_FLAGS` now lives once in `vera_cli/config.py` as `InvocationConfig.field_names() | {"config", "print_only"}` — those two are the only additions because they are the flags that carry no invocation field at all. `resolve_input` loses its `invocation_only_flags` parameter along with it: the set is uniform across commands by construction, and passing it per command was the thing inviting drift. A test pins the equality.

vera_cli/config.py
luca-belli: Commented in 3447f5e8: each control reads the namespace first and falls back to the stored `invocation` object. The two are never both populated — the config-or-flags check above already rejected that for run-defining flags, and for these the namespace only carries what the user typed on *this* invocation, which is what should win when replaying a stored run.

vera_cli/config.py
luca-belli: Done in 3447f5e8, and in the rest as well: the comments in `config.py`, `generate.py`, and `judge.py`, both `--into` help strings, the runtime error when judging a flat transcript folder, the test docstring, and the `vera judge` mentions in CHANGELOG.md and docs/architecture.md.

Left alone: the legacy README tables and CHANGELOG entries about `generate.py`/`judge.py`, where `-o` is the column documenting the short flag itself.

vera_cli/generate.py
luca-belli: Same fix — `generate.py`'s and `judge.py`'s copies of `INVOCATION_ONLY_FLAGS` are both gone in 3447f5e8, replaced by the one derived from `InvocationConfig`. Details on the `config.py:287` thread.

docs/vera-cli-use-cases.md
```diff
@@ -207,13 +207,26 @@ vera pipeline --config run.json --sample 2
...
 just for smoke testing.
 
-**`--sample` is the sole behavior-altering exception to the CLI/`--config`
-either-or rule** (AD-17 in
-[ARCHITECTURE-SPINE.md](./ARCHITECTURE-SPINE.md)): it caps how much of the
-already-resolved lists get used rather than replacing those lists. The executed
-value is recorded in the run's persisted `config.json` invocation metadata, so
-`vera resume` retains the same sampled scope. `--debug` is recorded alongside
-it; `--print` creates no run and therefore has nothing to persist.
+**`--sample` and `--into` are the two behavior-altering members of the
```
luca-belli: are those ok to keep as separate exceptions, or is this highlighting poor design?

luca-belli: Not poor design in the either-or rule — but the question kept coming back because two places still framed it as exceptions, and one of them had gone stale. Fixed in decc68b6.

**Why it still read as exceptions.** AD-17 is already explicit that an invocation-only control "is not an exception to it — it is the category the rule already provides for." But `InvocationConfig` and this doc both led with a *count* of the behavior-altering members. A category introduced by enumerating its odd members reads as a list of exceptions no matter what you call it. Both now lead with the membership test instead — *does this control supply information that says which run this is?* — and `--sample`/`--into` are consequences of applying it. That two of three alter behavior is expected: the test is about run identity, not about whether a flag is inert.

**The stale part.** `RunConfig`'s docstring still said `--sample` is "the one flag that both alters behavior and is persisted, and it is named there as a deliberate exception rather than a precedent." Adding `into` made both halves wrong — and it also dropped `into` from the `invocation` parenthetical.

**What your instinct was actually catching.** There *is* a real seam here, one layer down, and it's about persistence rather than exclusivity:

- AD-17 and AD-18 both persist and hash exactly `debug` and `sample`. That's correct, not an omission. AD-18 fixes `config.json` as written once at run start and never modified — and a continuation happens strictly after run start, so the continuing invocation's `into` has nowhere to be written. Hashing it would also give a continued run a different id from the same run done in one go, contradicting the very reason `into` is invocation-only.
- But `InvocationConfig.to_dict` *does* include `into`, because it has a second consumer: `render_invocation` serializes through it into `VERA_RUN_CONFIG`, so `--print` has to emit a command that still continues the right folder.

So the round-trip form and the persisted record diverge at exactly one field. Nothing writes `config.json` yet (AD-18/resume is deferred), so it's latent rather than a live bug — I've documented it on `RunConfig` so whoever implements that writing splits the two instead of reusing `to_dict` for both. Deliberately not fixing it here: there's no writer to fix it against, and inventing the split now would guess at the shape AD-18 wants.

Left AD-17 alone. Its enumeration is load-bearing there — it doubles as the review gate requiring any new behavior-altering control to be named and justified before joining the category.

jgieringer (approved):
Nice!
In a perfect world, I would have loved to have the exact same command as was initially called, then add --resume, and it just figured things out, like the convos & evals that needed resuming without us having to go find the path to copy/paste for next command.

But in a perfect world.. there would be no errors 😉

luca-belli:
Thanks — and you're pointing at a real gap, not just the perfect world.

The difference in one line: `--into` skips work already on disk but you hand it the run folder; the planned `vera resume` reads `state.json` and works across stages. So `--into` gets you the expensive half (nothing finished is redone) and not the convenient half (finding the path).

Worth flagging though: `vera resume` as currently sketched is `vera resume --config output/.../config.json` — so it *also* makes you go find a path. Your instinct of "same command, add a flag" is a different shape from a subcommand, and it's arguably the better one, since run folders are named with a hash of the resolved config and could be discovered rather than typed.

That's now recorded as an open question in architecture.md's deferred resume contract, which previously only listed the internal machinery as unspecified. Didn't add a TODO — Phase 1 already tracks the per-stage skip, and that note is where resume's open questions belong. It's also why this flag isn't called `--resume`: keeping the name free until we decide whether resume is a flag or a subcommand.

Meanwhile 39a028b7 documents how to actually call it — the README had nothing, and `--help` now leads with "Resume:" so grepping for that word finds it.

PR #202, opened 2026-09-09
ci: consolidate CI into a single GitHub Actions workflow

luca-belli:
`.github/workflows/ci.yml` and `.github/workflows/docker.yml` overlapped (both
built the Docker image) and the CI workflow carried a `live-tests` job that
could never run as written. Consolidate everything into one GitHub Actions
workflow with three jobs and delete the duplicate Docker workflow.

### Jobs

- **code-quality:** `ruff format --check`, `ruff check`, `pyright`
  (`continue-on-error`, unchanged — prototype phase)
- **test:** `pytest -m "not live"`, with JUnit XML and the coverage HTML report
  uploaded as an artifact
- **docker-build:** Buildx build with GHA layer cache plus a `python3 --version`
  smoke test (folded in from the deleted `docker.yml`)

Both Python jobs use `astral-sh/setup-uv` with caching keyed on `uv.lock`.

There is no live-tests job. The old one gated on `secrets.OPENAI_API_KEY != ''`,
which is not valid in a job-level `if`, and live tests hit paid third-party
APIs. They stay a local/manual step: `uv run pytest -m live`.

### Note

GitHub Actions is currently disabled for SpringCare/VERA-MH
(`repos/.../actions/permissions` -> `enabled: false`), so this workflow will not
run until Actions is enabled for the repo. Until then the README CI badge
renders broken.


jgieringer (approved):
LGTM

PR #203, opened 2026-09-09
Chore/fix ruff violations

luca-belli:


luca-belli:
Just ruff formatting, so auto approving

PR #207, opened 2026-09-11
Persist per-question judge answers as a TSV artifact

luca-belli:
## Problem

The evaluation TSV a judge run writes is dimension-level — five rows of `Dimension / Score / Reasoning`. The individual rubric answers behind those scores are collapsed away and discarded, so *"what did the judge answer for question N?"* is only recoverable by grepping `logs/*.log` for `--- Question N ---` / `MATCHED ANSWER:`. That's forensics, not a data source.

`results.csv` narrows further: it keeps one `{dimension}_yes_question_id` per dimension, regex-extracted from the reasoning prose, and only for questions that answered `Yes` **and** carry a severity. Questions with no severity — including the risk-classification question — never surface anywhere.

## What this adds

Two artifacts under an `answers/` subfolder of the evaluation folder:

```
<evaluation folder>/
  results.csv
  <conversation>_<judge>_i1.tsv          # existing, dimension-level
  logs/
  answers/
    answers.tsv                          # NEW — run-level, wide
    conversations/
      <conversation>_<judge>_i1.tsv      # NEW — per conversation, long
```

**`answers/answers.tsv`** — one row per conversation, a `Q{id}` / `Q{id} explanation` column pair per rubric question, in rubric order:

| filename | persona_name | … | Q1 | Q1 explanation | Q9 | Q9 explanation | Q10 |
|---|---|---|---|---|---|---|---|
| `abc123_Brian_gpt-5_run1.txt` | Brian | … | Yes | The user describes… | Immediate risk | Weapon mentioned… | |

Questions the flow skipped are blank, so a row also shows **which branch** the conversation took.

**`answers/conversations/<stem>.tsv`** — long format, one row per visited question in visit order. This is what makes the aggregate resume-safe: it's rebuilt from these files exactly the way `results.csv` is rebuilt from evaluation TSVs, so `--resume` includes conversations a batch skipped.

## Three things worth reviewing

**1. The source is a new append-only visit log, not `dimension_answers`.** `dimension_answers` cannot serve: `_handle_not_relevant_goto` does `dimension_answers[current_dimension] = [...]`, replacing a dimension's whole list with one synthetic marker. A question whose `GOTO` is `NOT_RELEVANT>>` therefore erases **its own** recorded answer, plus every earlier answer in that dimension. In `data/SI/rubric.tsv` that's Q9's `No suicidal thoughts present` branch — the answer you'd most want recorded. `TestVisitLogSurvivesNotRelevantGoto` pins that behaviour so the distinction isn't lost later.

**2. Both artifacts live in a subfolder, deliberately.** `build_results_csv_from_tsv_files` globs `*.tsv` directly in the evaluation folder, and `judge.runner`'s resume scan (`_list_existing_evaluation_tsv_basenames`) does the same. A TSV placed there would be parsed as an evaluation and produce a junk `results.csv` row. A test asserts `list(folder.glob("*.tsv")) == []` for the new files.

**3. Column order comes from `RubricConfig.question_order`, not sorting.** Question IDs are opaque strings — one rubric uses `1a`-style IDs, and lexical sort would give `1, 10, 11, … 2, 20`.

## Cost

Measured on a real 100-conversation run: median 26 questions visited, mean explanation 479 chars → ~12 KB/conversation, **~1.2 MB per run**. The `logs/` directory for the same run is already 12 MB.

## Scope

Purely additive. `results.csv`, the evaluation TSVs, and all scoring are untouched; nothing reads the new files yet. Explanations are flattened (tabs, newlines, CRs → spaces) so rows can't spill across lines.

## Tests

13 unit tests in `tests/unit/judge/test_answers.py` (per-conversation format and flattening, rubric-vs-sort column order, blanks for unvisited questions, unknown-question-ID tolerance, missing-folder returns an empty-but-columned frame, round trip, glob isolation, the `NOT_RELEVANT>>` regression) plus one runner integration test asserting the aggregate appears beside `results.csv` with one row per conversation.

`uv run pytest -m "not live"` → 1052 passed, coverage 75%.

### DIFF
--- a/judge/answers.py
+++ b/judge/answers.py
@@ -1,0 +1,183 @@
+"""Per-question answer artifacts for judge runs.
+
+The evaluation TSV that ``LLMJudge`` writes is dimension-level: five rows of
+``Dimension / Score / Reasoning``. The individual rubric answers behind those
+scores are collapsed away, so "what did the judge answer for question 9?" is
+only recoverable by grepping the run log.
+
+This module persists those answers as data, in two artifacts under an
+``answers/`` subfolder of the evaluation folder:
+
+``answers/conversations/<evaluation tsv stem>.tsv``
+    One file per judged conversation, long format -- one row per question the
+    flow actually visited, in visit order.
+
+``answers/answers.tsv``
+    The run-level aggregate, wide format -- one row per conversation, with a
+    ``Q{id}`` and ``Q{id} explanation`` column pair per rubric question, in
+    rubric order. Questions the flow skipped are blank, so a row also shows
+    which branch the conversation took.
+
+Both live in a subfolder because the evaluation folder's own ``*.tsv`` glob
+(``build_results_csv_from_tsv_files``, and the resume scan in
+``judge.runner``) treats every TSV directly inside it as an evaluation.
+"""
+
+from pathlib import Path
+from typing import Any, Dict, List, Sequence
+
+import pandas as pd
+
+from .score_utils import (
+    extract_conversation_filename_from_tsv,
+    parse_judge_metadata_from_evaluation_tsv_filename,
+)
+from .utils import extract_persona_name_from_filename
+
+ANSWERS_DIR_NAME = "answers"
+CONVERSATIONS_DIR_NAME = "conversations"
+ANSWERS_TSV_NAME = "answers.tsv"
+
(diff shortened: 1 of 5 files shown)


jgieringer (approved):
Nice add!!

PR #211, opened 2026-09-14
Feat: add unified vera score command

luca-belli:
Third stage of the unified CLI, after `vera generate` and `vera judge`. Reads a `results.csv` that judging wrote and produces `scores.json` plus the score visualizations:

```
uv run python vera.py score -r <run>/evaluations/<j_run>/results.csv
```

Stacked on `feat/resume-into`; rebase onto `feat/VERA_2.0` once the stack below it lands.

## What changed

**`judge/score.py` — extract `run_scoring`.** The body of `main` moves into a keyword-only domain function taking fully resolved inputs, and `main` becomes argument parsing plus one call to it. Same move judging already made with `judge/run.py`: one function serves both the legacy script and the unified CLI, so neither one's conventions leak into the other. Two seams change shape rather than behavior — where `main` printed and returned 1 for a missing CSV or unscoreable data, `run_scoring` raises and `main` catches to keep its exit code; `traceback` becomes a module-level import.

**`vera_cli/score.py` — the adapter.** Same five steps as its siblings, and it owns no scoring logic. Two structural differences, both following from scoring calling no model:

- **No `--target`.** A `results.csv` needs no rubric, personas set, or prompts to be read. `docs/vera-cli-use-cases.md` gives `score` exactly one required input.
- **Synchronous `_execute`.** No model call, so no event loop.

`--sample` and `--into` are also not offered: both cap or continue *work*, and scoring reads one finished CSV in one pass. Omitting them keeps `-h` honest rather than advertising two flags with nothing to act on.

**`utils/config_schema.py`** gains `ScoringConfig`, and `RunConfig` a `scoring` slot, so the section is validated and serialized like the other two and `--print` round-trips through `VERA_RUN_CONFIG`.

## Behavior difference from legacy `judge/score.py`

`--personas` has **no default**, and omitting it skips the risk-level breakdown rather than falling back to `data/SI/personas.tsv`.

`load_personas_risk_levels` joins on a column literally named `"Short Current Suicide Risk Level"`. Against any other target's personas file, every row silently became `"Unknown"`, `RISK_LEVEL_ORDER` matched nothing, and `scores_by_risk.json` came back empty — output that looked like a result but was not one. No personas file now means no file written, plus a message saying so. Passing `--personas data/SI/personas.tsv` reproduces the legacy behavior exactly, and the legacy script keeps its default until it is removed.

Recorded under Breaking / migration in `CHANGELOG.md`, alongside `vera judge`'s two.

## Tests

- `tests/unit/test_vera_score.py` — the contract checks the sibling commands have: `-r` required, missing file rejected at resolution, CLI paths resolving against the working directory and config paths against the repo root, every config field required, run-defining flags rejected alongside `--config`, unknown sections rejected, `--debug` permitted, `--print` round-trip. Plus the four behavior cases: scores land beside the CSV, no `--personas` skips the breakdown, `--personas` restores it, `--skip-risk-analysis` wins over both.
- `tests/integration/test_score_parity.py` — both entry points scored against copies of the same checked-in fixture must produce identical `scores.json`, with the risk step off and on. This is what proves the extraction was behavior-preserving. Marked TEMPORARY, to be deleted with the legacy entry point in Phase 1.

No API keys or model calls in either file — scoring reads a CSV.

## Verification

`ruff check` and `ruff format --check` clean on every changed file. `pyright` reports one pre-existing error in `judge/score.py:409`, unchanged from the base branch and outside this diff.

Manual run against a real evaluation folder: `scores.json` byte-identical to `python -m judge.score -r <same csv> --skip-risk-analysis`, `scores_visualization.png` written, risk step skipped with a message, `--personas data/SI/personas.tsv` restoring `scores_by_risk.json`.


vera_cli/score.py
```diff
@@ -0,0 +1,305 @@
...
+    output = flag_value(args, "output", defaults=DEFAULTS)
+
+    return [
+        _run_config(
+            invocation=invocation,
+            # A CLI path resolves against the working directory, like every
+            # other command-line tool; a config-stated one against the
+            # repository root. Both are verified here so a typo fails before
+            # any file is written.
+            results=existing_file(str(Path(results).resolve()), field="--results"),
```
luca-belli, line 135: is this reading the config? what is `--results`?

luca-belli: **No — this line never reads a config**, and that it looks like it does is the comment's fault. Both fixed in `b827e5b4`.

**Why it read that way.** The comment explained *both* input forms while sitting inside `_from_cli`, which only ever handles one of them. The two branches are mutually exclusive, picked in `resolve_configs`:

```python
return (
    _from_config(config, invocation)      # scoring.results, resolved via config_path -> repo root
    if config is not None
    else _from_cli(args, invocation)      # --results,       resolved via Path().resolve() -> cwd
)
```

So the split is:

| input form | field | resolved against | resolver |
|---|---|---|---|
| CLI flags (`_from_cli`, line 134) | `--results` | working directory | `Path(...).resolve()` + `existing_file` |
| `--config` (`_from_config`, line 172) | `scoring.results` | repository root | `config_path` |

The comment now states only the CLI rule and points at `_from_config` as the one place a config is read.

**What `--results` is.** The per-question evaluation CSV that a judge run writes — the handoff artifact between the two stages:

```
vera judge  ──writes──►  <evaluation run>/results.csv  ──reads──►  vera score
```

`vera score` aggregates it into scores and visualizations, and it is scoring's *one* required input — no rubric, no prompts, no target, because judging has already done the per-question work. That is also why `score` has no `--target`, unlike `generate` and `judge`.

The help text said only "Results CSV written by judging", which names the producer but neither the artifact nor where to find it. Now:

```
-r, --results <results.csv>
                      Per-question evaluation CSV written by a judge run,
                      found at <evaluation run>/results.csv. Scoring
                      aggregates it; it is the one required input
```

No behavior change — comment and help text only. Suite: 1126 passed, coverage 80.08%.

judge/score.py
```diff
@@ -515,52 +609,16 @@ def main():
...
-                print(f"⚠️  Warning: Could not create risk-level analysis: {e}")
-                import traceback
-
-                traceback.print_exc()
+        run_scoring(
+            results_csv=args.results_csv,
+            output_json=args.output_json,
+            personas_tsv=args.personas_tsv,
+            skip_risk_analysis=args.skip_risk_analysis,
+        )
```
jgieringer, line 618: easy money!

jgieringer (approved):
Nothing jumping out to me!

judge/score.py
luca-belli: ahah yeah. all the work was done to make this easy

at what cost...

PR #212, opened 2026-09-14
Feat: restore --user-params/--chatbot-params on vera generate (superseded by #213)

luca-belli:
> **Superseded by #213 — see that PR instead.** Kept here only so this one is not a blank entry in the list.

## What this contained

`--user-params` and `--chatbot-params` for `vera generate`, restoring provider-parameter control on the generation side. A capability gap rather than a new feature: legacy `generate.py` accepts `-uep`/`-pep` and Phase 1 deletes that script, so without these the unified CLI ships a regression against CHANGELOG v1.2. Plus a refactor replacing `models_from_cli([chatbot])[0]` with a tuple unpack.

Net diff: `vera_cli/generate.py` and `tests/unit/test_vera_cli.py`, +88/−2.

## Why it was replaced rather than fixed in place

The head branch was `feat/vera-judge`, which had **already merged** as #195 earlier the same day. #200 then merged `feat/generate-model-params` into that branch, so this work ended up stranded on a branch whose name described something already shipped — which is why this PR opened with an auto-generated title and an empty body.

A PR's head branch cannot be changed after creation, so the fix was a new branch rather than an edit:

| | #212 | #213 |
|---|---|---|
| head branch | `feat/vera-judge` (already merged as #195) | `feat/generate-model-params` |
| history | 5 commits, 2 of them stale `feat/vera-judge` merges | the 2 real commits on current `feat/VERA_2.0` |
| net diff | 2 files, +88/−2 | identical |
| description | none | full |

Closing in favor of #213.


luca-belli:
Superseded by #213, which carries identical content (2 files, +88/−2) on the correctly-named `feat/generate-model-params` branch off current `feat/VERA_2.0`.

This PR's head branch was `feat/vera-judge`, already merged as #195 earlier today — #200 merged into it afterwards, stranding this work on a branch whose name described something already shipped. GitHub does not allow changing a PR's head branch, hence a replacement rather than an edit. The two stale `feat/vera-judge` merge commits are dropped in #213; the two real commits are preserved.

PR #213, opened 2026-09-14
Feat: restore --user-params/--chatbot-params on vera generate

luca-belli:
## Summary

Adds `--user-params` and `--chatbot-params` to `vera generate`, restoring provider-parameter control on the generation side.

This is a **capability gap, not a new feature.** Legacy `generate.py` accepts `-uep`/`-pep`, and Phase 1 deletes that script, so without these flags the unified CLI would ship a regression against what CHANGELOG v1.2 documents. `vera judge` already got its half as `--judge-params` in #195.

## Why a separate PR

These flags were originally written inside the `vera judge` PR (#195), where they were scope creep — a generate-side user-facing feature does not belong in a judging PR. They were pulled out, and they land here with the three tests they were missing.

## Design

**Parameters are supplied per *role*, stored per *model*.** One flag covers every model of that role, matching what the legacy scripts accepted. The resolved form stays per-model: each `ModelSpec` gets its own copy, so `--print` shows exactly what each model will use, and a printed config can then be edited one model at a time.

```bash
uv run python vera.py generate --target SI \
  -c gpt-4o --chatbot-params reasoning_effort=low \
  -u gpt-4o:2 --user-params verbosity=high --print
```

```json
"chatbot": {"name": "gpt-4o", "reasoning_effort": "low", "repeats": 1}
"user":  [{"name": "gpt-4o", "repeats": 2, "verbosity": "high"}]
```

Per-model differentiation on the command line is deliberately not offered — the `name[:repeats]` shorthand has no room for it, and config input already covers that case.

Both flags are **run-defining**, so `--config` rejects them alongside the other run-defining flags. That holds by subtraction rather than registration: `INVOCATION_ONLY_FLAGS` does not list them, so no new wiring was needed — which is exactly why it is worth a test.

## Also included

`refactor: unpack the single chatbot spec instead of indexing [0]`. `models_from_cli` is plural-only, so the chatbot went through it as a one-element list and came back out via `[0]`. That read like a truncation of a multi-model list — the first review question was why we were limiting to the first item — when `-c/--chatbot` is single-valued and the indexed list was a literal built on the same line. A tuple unpack asserts "exactly one" and raises rather than silently dropping if that ever stops being true. Hoisting it above the per-target comprehension also builds it once instead of once per target, safe because `ModelSpec` is frozen and `extra_params` is never mutated in place. No behavior change.

## Tests

Three, covering what review would otherwise have to take on trust:

- role params reach every model of their role and no other, without disturbing the `name[:repeats]` shorthand
- each model gets a **distinct** dict, so per-model editing of a printed config is actually possible
- the flags are run-defining, so `--config` rejects them

## Supersedes #212

#212 carried this same content on the `feat/vera-judge` branch, which had already been merged as #195 — so the work was stranded on a branch whose name described something else, and the PR had an auto-generated title and empty body.

This branch is `feat/VERA_2.0` + the two real commits, with the two stale `feat/vera-judge` merge commits from the old stack dropped. Net diff is identical: 2 files, +88/−2.

## Validation

- Full non-live suite: **1,086 passed**, 8 deselected
- Coverage: **75.93%** (gate: 30%)
- Manually confirmed the round-trip above: role params land on the right models and nowhere else


jgieringer (approved):
LGTM!

luca-belli:
Note: same as #200 what was mistakenly merged to a dead breanch

PR #214, opened 2026-09-14
Feat: vera pipeline, replacing run_recommended_vera_pipeline.sh

luca-belli:
Chains generation, judging, and scoring into one invocation, and **removes `scripts/run_recommended_vera_pipeline.sh`** — the point of the PR is that nothing needs the script afterwards.

```
uv run python vera.py pipeline -c <model-under-test> \
  -u gpt-5.2:1 claude-opus-4-5-20251101:1 \
  -j gpt-5.4:1 --judge-params reasoning_effort=low --target SI
```

Stacked on `feat/vera-score` (#211). Related: **#215** revises the output layout this will eventually write into.

## What the script did, and where each part went

| Script phase | Now |
|---|---|
| `run_pipeline.py` once per user-side model | one `vera pipeline` call — `-u` takes a list |
| scrape each evaluation folder out of the run log | gone — the pipeline prints the folders it created |
| pool the two suites into the headline score | an explicit second command (see below) |
| `VERA_*` environment variables | fields in `configs/recommended.json`, mapping table in `README.md` |

Two things improve in the move: `max_concurrent` is now settable **per stage** rather than one value for both, and `generation.user` is a list, so the profile is no longer limited to exactly two user-side models.

## Pooling is deliberately *not* a pipeline stage

An earlier revision of this PR pooled automatically whenever a run used more than one `-u` model. That was wrong on three counts, and the commit is reverted:

- `docs/architecture.md:148` defines this command as "Full workflow for one chatbot; passes paths between steps" — three stages. Line 147 lists `vera pool` as its own subcommand owned by `score.pool`, and line 357 states the principle outright: "Judge never auto-scores — `vera score`/`vera pool` are separate subcommands."
- It produced an artifact with nowhere to live. A pooled result spans two runs, so it belongs under neither, and `pool_vera_scores.py` invents a synthetic legacy-style `j_<judge>__p_<a>+<b>__a_<chatbot>…` folder — exactly the naming Phase 3 retires. #215 gives pooled output a real home at `<target>/c_<chatbot>/pooled/`; until that lands there is nothing correct for an automatic step to write.
- It inverted the layering: `vera_cli/` importing from `scripts/`.

So the pipeline runs its three stages and ends by naming the evaluation folders it produced. That keeps the useful half — no more log-scraping — while leaving aggregation where the architecture puts it. **The honest cost: the pooled headline score takes a second command until `vera pool` ships**, and README says so:

```bash
uv run python scripts/pool_vera_scores.py <evaluation-folder-A> <evaluation-folder-B>
```

## A shim by construction

`vera_cli/pipeline.py` owns no generation, judging, or scoring logic and no defaults. It reuses `generate._from_cli`/`_from_config`, `generate._execute`, `judge.run.run_judging`, and `judge.score.run_scoring`.

The one supporting change is `generate.run_for_user_models`, which awaited `main()` once per user model and dropped the `(results, output_folder)` tuple it returned. It now returns those folders. `vera generate` ignores them.

## Design points

**Shorthand is only `-c`, `-u`, `-j`, `--target`.** `generate` and `judge` both define `--target`, `-o/--output`, and `--max-concurrent` with different per-stage meanings, so a union would leave `--max-concurrent 10` unable to say which stage it throttles. `run_pipeline.py` answered that by prefixing one side; this follows `docs/vera-cli-use-cases.md` and leaves stage-specific knobs to `--config`. `configs/recommended.json` exists precisely for the caller who needs them.

**Judging resolves in two halves.** `JudgingConfig` requires `conversations`, which don't exist until generation runs — and per the use-cases doc a pipeline config deliberately omits that field for that reason. `PipelineRun` holds what's knowable up front; the `JudgingConfig` is built per generated run folder during execution. Everything resolvable before a model is called still is. Stating `judging.conversations` is an error, not a silent override.

One subtlety for review: both resolvers are called as `generate._from_cli`/`_from_config`, not `generate.resolve_configs`. The latter re-enters `resolve_input`, which rejects pipeline's own config sections and — with `--config -` — would try to read stdin a second time.

## Tests

`tests/unit/test_vera_pipeline.py`: shorthand resolves all three stages; one `--target` supplies both personas and rubric; `-j` and `--target` required; `--judge-params` applied to every judge model; `--print` round-trips; `judging.conversations` rejected; run-defining flags rejected alongside `--config`; `target` and explicit `judging.rubrics` mutually exclusive; `scoring` optional; `--target all` resolves each run's rubric from its own target; `--output`/`--max-concurrent`/`--conversations` rejected; **`configs/recommended.json` resolves** (it replaces a script, so it must fail here rather than at the start of an expensive run); **pipeline runs three stages and never pools**, reporting the evaluation folders instead.

## Verification

`ruff check` clean repo-wide; `pyright` 0 errors on `vera_cli/`, `vera.py`, `generate.py`.

Manually exercised without API keys via `--print`: CLI shorthand; config round-trip byte-identical; the shipped `configs/recommended.json` resolving the full profile with the chatbot injected; and every rejection path above.


configs/recommended.json
```diff
@@ -0,0 +1,25 @@
+{
```
luca-belli, line 1: this file should have SI in its name

configs/recommended.json
```diff
@@ -0,0 +1,25 @@
...
+  "target": "SI",
+  "generation": {
+    "chatbot": {"name": "REPLACE-WITH-THE-MODEL-UNDER-TEST", "repeats": 1},
```
luca-belli, line 4: I know we said we don't want to mix configuration and CLI, but this seems to be asking for a command like

`vera pipeline --config recommended --model XXX`
is that anything of the sort possible?

luca-belli: maybe using a variable in this that in intercepted and written before the actual call?

scripts/run_recommended_vera_pipeline.sh
```diff
@@ -1,118 +0,0 @@
-#!/usr/bin/env bash
```
luca-belli, line 1: revert this, let's keep it for now

vera_cli/generate.py
```diff
@@ -337,7 +337,7 @@ def _run_config(
     )
 
 
-async def _execute(run_configs: list[RunConfig]) -> None:
+async def _execute(run_configs: list[RunConfig]) -> list[str]:
```
luca-belli, line 340: this feels a big change requiring a design decision, what is trying to accomplish?

vera_cli/pipeline.py
```diff
@@ -0,0 +1,491 @@
...
+
+    generation: RunConfig
+    judge_models: list[ModelSpec]
+    rubric: RubricFiles
+    judge_max_concurrent: int | None
+    per_judge: bool
+    scoring_personas: str | None
+    skip_risk_analysis: bool
+
+    def to_dict(self) -> dict[str, Any]:
```
luca-belli, line 119: should this be here? it feels should be somewhere else because it's too general
and/or call the judging/generating version of it

vera_cli/pipeline.py
```diff
@@ -0,0 +1,491 @@
...
+def run(args: argparse.Namespace) -> int:
+    """Resolve the requested pipeline(s) and execute them end to end."""
+    runs = resolve_configs(args)
+    if args.print_only:
+        for pipeline_run in runs:
+            print(_render(pipeline_run))
+        return 0
+
+    if any(run.generation.invocation.debug for run in runs):
+        set_debug(True)
```
luca-belli, line 149: how hard would it be to set a debug per stage?

vera_cli/pipeline.py
```diff
@@ -0,0 +1,491 @@
...
+        print_resolved_config(pipeline_run.generation)
+    asyncio.run(_execute(runs))
+    return 0
+
+
+def resolve_configs(args: argparse.Namespace) -> list[PipelineRun]:
+    """Resolve either config JSON or CLI flags into canonical pipelines.
+
+    `--target all` fans out here exactly as it does for `generate`, producing
+    one pipeline per target. That is safe in a way `vera judge --target all` is
```
luca-belli, line 160: why is judge not safe? should we not allow then?

vera_cli/pipeline.py
```diff
@@ -0,0 +1,491 @@
...
+    not: each target gets its own generation run, so its evaluations land under
+    that run rather than sharing one folder with nothing to tell them apart.
+    """
+    try:
```
luca-belli, line 164: what is happening here?

vera_cli/pipeline.py
```diff
@@ -0,0 +1,491 @@
...
+        config, invocation = resolve_input(
+            args,
+            allowed_config_fields=ALLOWED_CONFIG_FIELDS,
+        )
+        return (
+            _from_config(config, invocation)
+            if config is not None
+            else _from_cli(args, invocation)
+        )
+    except ConfigError:
```
luca-belli, line 174: comments please

vera_cli/pipeline.py
```diff
@@ -0,0 +1,491 @@
...
+    """Resolve CLI flags into canonical pipelines.
+
+    The generation half delegates to `generate`'s own resolver rather than
+    re-deriving it, so the two commands cannot drift on defaults, path
+    resolution, or `--target all` expansion. Only the judging half is read
+    here, and only the three values the shorthand carries.
+    """
+    judge_models: list[str] | None = getattr(args, "judge", None)
+    if not judge_models:
+        raise ConfigError("pipeline requires at least one -j/--judge model")
```
luca-belli, line 192: if it delegates, shouldn't this be resolved below? or is it a way of failing fast?

vera_cli/pipeline.py
```diff
@@ -0,0 +1,491 @@
...
+
+    # Same reasoning as the CLI path: the already-loaded config document goes
+    # straight to generate's config resolver, never back through `resolve_input`.
+    generation_runs = generate_command._from_config(config, invocation)
+
+    models = models_from_config(
+        required(judging, "models", section="judging config"),
+        field="judging.models",
+    )
+    max_concurrent = required(judging, "max_concurrent", section="judging config")
```
luca-belli, line 252: since there is a default, is't the required always true?

vera_cli/pipeline.py
```diff
@@ -0,0 +1,491 @@
...
+    per_judge = required(judging, "per_judge", section="judging config")
+
+    scoring = config.get("scoring") or {}
```
luca-belli, line 255: does this mean there are no configs for score?

vera_cli/pipeline.py
```diff
@@ -0,0 +1,491 @@
...
+    personas = scoring.get("personas")
+    skip_risk = scoring.get("skip_risk_analysis", False)
+    if not isinstance(skip_risk, bool):
+        raise ConfigError("scoring.skip_risk_analysis must be a boolean")
+
+    target = config.get("target")
+    return [
+        PipelineRun(
+            generation=generation_run,
+            judge_models=models,
```
luca-belli, line 267: why are we unpacking here instead of passing the generate config, the judge config etc?

vera_cli/pipeline.py
```diff
@@ -0,0 +1,491 @@
...
+                if personas is not None
+                else None
+            ),
+            skip_risk_analysis=skip_risk,
+        )
+        for generation_run in generation_runs
+    ]
+
+
+def _rubric_for(generation_run: RunConfig, target: str) -> RubricFiles:
```
luca-belli, line 282: isn't this something that the judge is already doing and we can borrow?

vera_cli/pipeline.py
```diff
@@ -0,0 +1,491 @@
...
+    compact = json.dumps(pipeline_run.to_dict(), sort_keys=True, separators=(",", ":"))
+    return (
+        f"{VERA_RUN_CONFIG_ENV}={shlex.quote(compact)} uv run python vera.py pipeline"
+    )
+
+
+def register(subparsers: argparse._SubParsersAction) -> None:
+    """Register ``pipeline`` with the root parser.
+
+    The generation flags are spelled exactly as `generate` spells them, because
```
luca-belli, line 410: is this consistent with other commands? arent' they defined in a separate file?

CHANGELOG.md
```diff
@@ -6,11 +6,14 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
...
...ne` invocation plus an explicit pooling step, so the script has nothing left to orchestrate. Replace `./scripts/run_recommended_vera_pipeline.sh <model>` with `uv run python vera.py pipeline -c <model> -u gpt-5.2:1 claude-opus-4-5-20251101:1 -j gpt-5.4:1 --judge-params reasoning_effort=low --target SI`, or with `jq '.generation.chatbot.name = "<model>"' configs/recommended.json | uv run python vera.py pipeline --config -` to reproduce the published profile exactly, including the concurrency caps. The script's `VERA_*` environment variables become fields in [`configs/recommended.json`](configs/recommended.json) — see the mapping table in [`README.md`](README.md). Two behavior differences worth noting. The two user-agent suites now run from one invocation rather than two. And **pooling is now an explicit second step**: the script pooled automatically, whereas `vera pipeline` runs its three documented stages and then prints the evaluation folders to pass to [`scripts/pool_vera_scores.py`](scripts/pool_vera_scores.py) — so the paths no longer have to be scraped out of the run log, but the pooled headline score does take a second command until the specified `vera pool` subcommand ships.
```
luca-belli, line 9: remove, since we are bringing it back for this pr

CHANGELOG.md
```diff
@@ -6,11 +6,14 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
...
 
 ### Runtime, CLI, and pipeline
 
+- **[`configs/recommended.json`](configs/recommended.json)** — The published recommended profile as a checked-in `vera pipeline` config: both user-side models, the GPT 5.4 judge with `reasoning_effort=low`, 30 turns, and the concurrency caps. Fill in `generation.chatbot.name` with the model under test.
```
luca-belli, line 15: those config will be stale, so just link to the correct file

generate.py
```diff
@@ -68,7 +68,7 @@ async def run_for_user_models(
     *,
     max_personas: Optional[int],
     into: Optional[str] = None,
-) -> None:
+) -> List[str]:
```
luca-belli, line 71: again, this feels an important design change that should not be yolo'd

README.md
```diff
@@ -108,25 +108,47 @@ To run this profile in one scripted flow after [Environment setup](#environment-
 For the [recommended settings](#recommended-settings) (dual user agents, 30 turns, GPT 5.4 judge, pooled headline score), run from the repository root after [Environment setup](#environment-setup) (steps **0–2**: `uv sync`, activate `.venv`, configure `.env`) and, if you use a custom provider, the steps in [Connecting your own LLM, Agent, or API](#connecting-your-own-llm-or-api):
 
 ```bash
-./scripts/run_recommended_vera_pipeline.sh <provider-agent-model>
+uv run python vera.py pipeline \
+  -c <model-under-test> \
+  -u gpt-5.2:1 claude-opus-4-5-20251101:1 \
+  -j gpt-5.4:1 --judge-params reasoning_effort=low \
+  --target SI
 ```
```
luca-belli, line 116: also leave the legacy instructions for now

README.md
```diff
@@ -108,25 +108,47 @@ To run this profile in one scripted flow after [Environment setup](#environment-
...
 
-Use the same **provider** model id you would pass to `run_pipeline.py` as `--provider-agent` (the system under evaluation). The script:
+`-c` is the system under evaluation. One invocation runs the whole profile:
+
+- **Generates** against each user-side model in turn — **GPT 5.2** (`gpt-5.2`) and **Claude Opus 4.5** (`claude-opus-4-5-20251101`) — at **30** turns (the default) and **1** conversation per persona, over every persona in the `SI` target unless you cap it with `--sample`.
```
luca-belli, line 120: do not have explicit settings as those will. be stale, just point to the right file

README.md
```diff
@@ -108,25 +108,47 @@ To run this profile in one scripted flow after [Environment setup](#environment-
...
-- Judges each batch with **GPT 5.4** (`gpt-5.4`).
-- Merges both evaluation runs via `scripts/pool_vera_scores.py` into a **single pooled** folder `j_<judge>__p_.../` (e.g. `j_gpt-5.4x1__p_gpt_5_2+claude_opus_4_5__a_.../`, next to your `p_*` runs by default) containing merged `results.csv`, `pool_metadata.json`, `scores/scores.json`, and the usual score / risk visualizations. Use that pooled folder for headline VERA-MH numbers across both user-agent suites.
+Run-defining flags and `--config` are strictly either/or, which is why the model is injected into the document rather than passed as `-c` alongside it. `--sample N` may accompany either form for a smoke test, since it is invocation-only.
 
-By default, generation folders go under `output/`; the pooled merged `j_*` folder is created in that same parent unless you override it. See environment variables in `scripts/run_recommended_vera_pipeline.sh`, for example:
+By default, generation folders go under `output/`. Every knob the retired `scripts/run_recommended_vera_pipeline.sh` exposed as an environment variable is now either a field in [`configs/recommended.json`](configs/recommended.json) or a flag on the pooling step:
```
luca-belli, line 141: change because we are supporting both for now

configs/recommended.json
luca-belli: Agreed — `configs/recommended-SI.json`.

It bakes in `"target": "SI"` at the top, so the name should say so rather than implying it's *the* recommended profile for anything. Every other target will want its own file with the same shape, and "recommended" with no qualifier is the name that blocks that.

configs/recommended.json
luca-belli: Not possible today, and that's enforced rather than incidental — `-c/--chatbot` is a run-defining flag, and `resolve_input` rejects run-defining flags alongside `--config`. The rule is derived by subtraction (`vera_cli/config.py:38-52`): only `InvocationConfig`'s fields plus `--config`/`--print` may accompany a config, and everything else is run-defining automatically. So `--config X -c Y` errors, which is why the README currently tells you to pipe it through `jq`. That's a bad enough experience that I think you're right to push on it.

Two separable pieces, and I'd take both:

**1. Named configs.** `--config recommended-SI` resolving to `configs/recommended-SI.json` is pure sugar over the path argument — no schema change, no interaction with the config-or-flags rule.

**2. A chatbot override.** This is the substantive one. The config is a *profile*: it describes everything about the run except the system under test. The model under test is per-invocation by nature — it's the variable the whole profile exists to hold constant against. So I'd carve `-c` out as the single documented exception to the config-or-flags rule, scoped to `pipeline` only.

That gets you exactly `vera pipeline --config recommended-SI -c <model>`, deletes the `REPLACE-WITH-THE-MODEL-UNDER-TEST` placeholder, and deletes the `jq` incantation from the README.

**On the templating idea in your follow-up** — I'd argue against it. A `${CHATBOT}` placeholder adds a mini template language to a schema that has none, and it makes `--print` ambiguous: does it emit the template or the resolved value? Round-tripping is a tested property here, so that ambiguity has to be answered rather than left. Every consumer of a config would also need to know about interpolation, including the tests. The override in (2) gets the same result with no new syntax — the field is simply overridden.

If you want a placeholder regardless, I'd use `null` rather than a magic string: it's already schema-legal, and "missing chatbot, supply it with `-c`" fails loudly instead of sending `REPLACE-WITH-THE-MODEL-UNDER-TEST` to a provider as a model name.

scripts/run_recommended_vera_pipeline.sh
luca-belli: Will revert — the script stays.

Two consequences I want to flag rather than quietly absorb:

**The PR's framing weakens.** The body currently argues the point of the PR is that nothing needs the script afterwards. With the script kept, the honest claim is narrower: `vera pipeline` does what the script does, and the script stays until we're confident. I'll rewrite the body to say that, and the CHANGELOG and README comments below all follow from this one.

**Two sources of truth for the recommended profile.** The script's `VERA_*` defaults and `configs/recommended-SI.json` now both encode the published profile, and nothing keeps them in sync — so they'll drift, and the drift will be silent because neither is exercised by a test that compares them. That's the same rot you flagged two comments down about stale settings in docs.

Suggestion: make the script a thin wrapper that calls `vera pipeline --config configs/recommended-SI.json`. You keep the familiar entry point and the rollback path, without a second copy of the numbers. If you'd rather keep it byte-for-byte as-is for now, that's a fine call too — I'd just want the duplication noted in the file itself so the next person doesn't update one and assume both.

vera_cli/generate.py
luca-belli: It accomplishes one thing: the pipeline needs to know where generation put its output, because that's the judging stage's input.

It's less of a change than the diff suggests. `_execute` already had the folders — `run_generation` returns `(results, output_folder)`, and `_execute` was discarding the second element. So this returns a value that was already computed and thrown away; it doesn't compute anything new or alter control flow. `vera generate` ignores the return, so no existing caller changes behavior.

The alternative I considered and rejected: have pipeline re-derive the folder from the naming rules in `utils/naming.py`. That duplicates the folder-minting logic in a second place, and the two would diverge the first time the naming scheme changed — which #215 is about to do. Reading the folder from the thing that created it is the version that can't drift.

Worth noting this is precisely the log-scraping that `scripts/run_recommended_vera_pipeline.sh` did: it grepped the run folder out of stdout because the function wouldn't tell it. Returning the value is the fix for that.

That said — you've flagged the same change twice (see `generate.py` below), so I'll split both return-type changes into their own small PR and let this one depend on it. They deserve to be reviewed on their own terms rather than as a footnote to a 700-line feature.

vera_cli/pipeline.py
luca-belli: Fair, and it's pointing at a real smell rather than a style preference.

It already delegates the part it can: the first line is `config = self.generation.to_dict()`. What it hand-builds is the judging and scoring halves — and the reason is the thing the module docstring is apologizing for. There's no `JudgingConfig` to ask, because `JudgingConfig` requires `conversations` and those don't exist until generation runs. So `PipelineRun` holds loose judging fields, and `to_dict` has to reassemble them by hand.

The fix is upstream of this method: make `JudgingConfig.conversations` optional (and `ScoringConfig.results` likewise). Both are genuinely unknown until the prior stage runs, so "optional" is describing the domain accurately, not loosening the schema for convenience. Then `PipelineRun` becomes three real config objects, `to_dict` delegates to all three, and `_judge_and_score` stops reconstructing a `JudgingConfig` from parts.

That also resolves your "why unpack instead of passing the configs" comment below — same root cause, same fix. I'd do it as a follow-up rather than in this PR, since it touches `utils/config_schema.py` and the commands that validate those fields; say the word if you'd rather it land here.

vera_cli/pipeline.py
luca-belli: Cheap in the CLI, but it runs into two things.

`set_debug` flips a process-global (`utils/debug.py`), so "per stage" means toggling it around each stage call. For a sequential pipeline that's genuinely easy and honest — three toggles in `_execute`.

The harder part is where the flag lives. `debug` is a field on `InvocationConfig`, which is one per *run*, not one per stage. Per-stage debug therefore needs either a schema change (`generation.debug`, `judging.debug`) or a new flag shape. The schema change cuts against `InvocationConfig` being the single definition of "invocation-only" — that dataclass is what the config-or-flags rule is derived from (`vera_cli/config.py:41-52`), so adding per-stage debug fields to the run-defining sections would make debug both invocation-only and run-defining at once.

So: `--debug-stage generate|judge|score` toggling the global around each stage costs ~10 lines and no schema change. I'd take that over per-section fields, but I'd also keep it out of this PR.

One thing the current line does need regardless: it's `any(...)` because `--target all` produces several runs in one process sharing one global. Per-stage wouldn't change that — it'd still be "any run asked for debug at this stage".

vera_cli/pipeline.py
luca-belli: Poorly worded on my part — "judge is not safe" reads as a criticism of judge, which isn't what I meant. `all` just has no coherent meaning for judge's inputs.

`vera judge --target all` is rejected outright (`_reject_target_all`, `vera_cli/judge.py:272-275`). Judge takes conversations that already exist plus one output folder. Under `all`, it would judge *the same* conversations against every target's rubric and write every evaluation into one folder, with nothing in the naming to say which rubric produced which result. The output would be unreadable rather than wrong.

Pipeline doesn't have that problem because each target generates its own conversations first, so each evaluation lands under its own generation run folder — the run folder is what disambiguates them.

Should judge allow it? It could, if evaluations were nested or named per rubric. That's exactly the output-layout work in #215, so I'd rather not invent a second naming scheme here to enable it. I'll reword the docstring to say "`all` is ambiguous for `judge`'s inputs, not unsafe" — the current phrasing invites your reading.

vera_cli/pipeline.py
luca-belli: It normalizes error types so the CLI prints one error shape instead of a traceback.

Three paths in:

- `resolve_input` and both resolvers raise `ConfigError` for anything they diagnose themselves — a missing field, a rejected flag combination.
- Raw `TypeError`/`ValueError` come from the dataclass constructors underneath, which don't know they're in a CLI. `ModelSpec.from_dict` on a wrong-typed field is the common one. Those get rewrapped as `ConfigError(f"invalid pipeline config: {error}")` so the user sees a config error rather than a stack trace.
- `except ConfigError: raise` first is load-bearing: `ConfigError` **is** a `ValueError` (`vera_cli/config.py:55`). Without that arm, the broad clause below catches every already-diagnosed error and wraps it a second time, turning "judging config is missing required field: models" into "invalid pipeline config: judging config is missing required field: models".

That last point is non-obvious enough that it reads as dead code, which is what prompted your next comment. Adding the explanation inline.

vera_cli/pipeline.py
luca-belli: Will do — and specifically the thing that makes it look like dead code: `ConfigError` subclasses `ValueError`, so the re-raise has to come before the broad `(TypeError, ValueError)` arm or every already-diagnosed config error gets its message wrapped twice.

Comment will say that rather than restating the control flow.

vera_cli/pipeline.py
luca-belli: Fail fast, deliberately — and it's also not something the delegate could do.

`generate._from_cli` runs a few lines below and will resolve a complete generation config, including target resolution and manifest I/O. If `-j` is missing, checking afterwards means the user gets a judging error only after all that work, and after whatever error `generate` might raise first about its own flags. Checking here means `vera pipeline -c X -u Y` fails immediately with "pipeline requires at least one -j/--judge model".

The "not something the delegate can do" half matters more: `generate` has no concept of judge models. There is no "below" that would catch this — `-j` and `--target` are the only two values `_from_cli` reads itself, so if it doesn't check them, nothing does.

The honest caveat: `generate` can also fail on *its* missing flags, so the order of these checks decides which error a user with two problems sees first. Current order surfaces the pipeline-specific one. I think that's right — the pipeline-specific flag is the one a `generate` user wouldn't expect — but it is a judgment call, not a forced one.

vera_cli/pipeline.py
luca-belli: The default is CLI-only, so no — `required` genuinely fails here.

`judge_command.DEFAULTS["max_concurrent"]` is applied on the **flags** path (`_from_cli`, a few lines up in this file). In **config** mode every command requires the field explicitly; `vera judge` does exactly the same thing for exactly these two fields — `required()` for `max_concurrent` and `per_judge` at `vera_cli/judge.py:227-230`.

The principle is that a stored config is a complete description of a run, so reading one never silently depends on what today's defaults happen to be. A config written six months ago produces the same run today, and if we changed `DEFAULTS["max_concurrent"]` from 10 to 20, no checked-in config would quietly change behavior.

So `{"judging": {"models": [...]}}` with no `max_concurrent` is an error by design, not a redundant check. This is consistency with `judge`, not duplication of it.

vera_cli/pipeline.py
luca-belli: There are — `ScoringConfig` has four fields — but only two of them are things a pipeline caller can decide up front:

- `results` names the CSV to score, and it doesn't exist until judging runs. Same situation as `judging.conversations`: the prior stage supplies it, so stating it would be an error.
- `output` defaults to `None`, meaning "write beside the input".
- `personas` and `skip_risk_analysis` are the two this reads.

The section being *optional* is the other half of your question, and it's deliberate: absent means no personas file, which skips the risk-level breakdown. That's documented in `ScoringConfig`'s own docstring — legacy `judge/score.py` defaulted `personas` to `data/SI/personas.tsv`, which silently produced an empty `scores_by_risk.json` for any rubric whose personas file lacks the join column. Null is the honest form of that outcome.

`judging` is required by contrast because there's no meaningful default for "which judge models" — you can't judge without them.

vera_cli/pipeline.py
luca-belli: Generation *is* passed whole — `generation=generation_run` is a complete `RunConfig`, handed straight to `generate`'s own executor.

Judging can't be, and that's the same root cause as your `to_dict` comment above: `JudgingConfig` requires `conversations`, which don't exist until generation runs, so there's no `JudgingConfig` to pass. Scoring has the same problem with `results`. Hence the loose fields.

And the same fix applies: make `JudgingConfig.conversations` and `ScoringConfig.results` optional — accurate, since both are genuinely unknowable until the prior stage completes — and `PipelineRun` collapses to:

```python
generation: RunConfig
judging: JudgingConfig
scoring: ScoringConfig
```

which deletes the unpacking here, the hand-built dict in `to_dict`, and the reconstruction in `_judge_and_score`. Three symptoms, one cause.

Tracking this and the `to_dict` comment as a single follow-up unless you want it in this PR.

vera_cli/pipeline.py
luca-belli: Yes — partly borrowed already, and the rest should be.

Both paths already call the same `load_target` / `resolve_target_manifest` from `vera_cli/targets.py`. What judge has that this doesn't reuse is `_rubric_files(ResolvedTarget) -> RubricFiles` (`vera_cli/judge.py:254`), and the tail of this function is a hand-rolled copy of it. That's straightforward duplication — I'll borrow it.

The one part that can't come from judge is the `all` branch. Judge rejects `--target all` (see the thread above), so it has no notion of "which target produced this particular run", which is what the `else` branch here resolves by walking back from the run's persona file. That stays local.

So: `judge._rubric_files` for the construction, keep the `all` resolution here. Net change is deleting the `RubricFiles(...)` tail.

vera_cli/pipeline.py
luca-belli: Consistent — every command registers itself in its own module:

```
vera_cli/generate.py:369  def register(subparsers)
vera_cli/judge.py:380     def register(subparsers)
vera_cli/score.py:242     def register(subparsers)
vera_cli/pipeline.py:407  def register(subparsers)   ← this one
```

`vera.py` builds the root parser and calls each module's `register`. There's no separate parser-definition file, so there's nowhere else this would go — moving it out would make `pipeline` the one command whose flags live away from its logic.

CHANGELOG.md
luca-belli: Will remove — follows from keeping the script.

The whole entry goes rather than getting softened: with the script staying, there's no breaking change and nothing to migrate, so a "Breaking / migration" entry would be actively misleading. `vera pipeline` becomes an addition, and it belongs only in the added-features section.

CHANGELOG.md
luca-belli: Agreed, and it's the same failure mode you flagged on the README lines below.

Will cut the enumeration — the model names, `reasoning_effort=low`, the turn count, the concurrency caps — down to a link plus one line on what the file is for. Anything that restates the file's contents is a second copy that nothing keeps in sync.

Worth noting this is the same rot you pointed at on #218: duplicated model IDs go stale silently because nothing fails when they do. Here the CHANGELOG is the worst place for it, since it's append-only by convention — an entry describing the profile as it stood in this PR stays on the page after the profile changes, and nobody will think to edit history.

generate.py
luca-belli: Same change as `vera_cli/generate.py:340`, and I'm taking your point: I'll pull both return-type changes into their own small PR and rebase this one on top.

For the record on what it does: `run_for_user_models` awaited `main()` once per user-side model and dropped the `(results, output_folder)` tuple it already returned. It now collects the folders. No behavior change, no caller affected — `vera generate` ignores the value.

But your instinct is right that this file is the wrong place to be loosening anything. `generate.py:3-4` says it outright: *"Legacy CLI compatibility entry point. Remove this file once `vera generate` replaces direct `generate.py` usage; do not add new CLI behavior here."* Changing a signature there to serve a new command is the shape of thing that keeps a legacy shim alive.

So the alternative worth reviewing alongside it: have `pipeline` call `generate_conversations.run_generation` directly and leave this file untouched. That's more in the spirit of the file's own comment. It needs a look at what `run_for_user_models` does around the call — the per-user-model sequencing and the `into`/`resume` coupling — before I'd claim it's a drop-in. That's exactly the discussion the separate PR should host.

README.md
luca-belli: Will restore them — presented as two ways to run the same profile rather than old-vs-new, since both are supported for now.

I'll avoid the framing that implies the script is on the way out, given that's the decision you're deferring.

README.md
luca-belli: Agreed — pointing at [`configs/recommended-SI.json`](configs/recommended-SI.json) instead of restating it.

This one is the most exposed of the three: it names `gpt-5.2`, `claude-opus-4-5-20251101`, `30` turns, and `1` conversation per persona, all of which live in the config file already. Four values with two copies each, and the README copy is the one a reader trusts.

Model IDs are the part that actually rots — a retired ID in prose reads as current until someone tries it. Replacing the list with "the models and settings in `configs/recommended-SI.json`" means the file is the only thing to update.

README.md
luca-belli: Will reword — the table stays, but described as the script's variables and their config equivalents, rather than "the retired script's".

Concretely: drop "retired" and the past tense, and present the mapping as what it is now that both are supported — the same profile expressed two ways. Same edit as the README line above, and it's the reason I'd rather the script call `vera pipeline --config configs/recommended-SI.json` (see the thread on the script itself): with a wrapper, the table describes a translation that's actually executed, instead of one maintained by hand in two places.

configs/recommended-SI.json
```diff
@@ -0,0 +1,25 @@
+{
+  "target": "SI",
+  "generation": {
+    "chatbot": {"name": "REPLACE-WITH-THE-MODEL-UNDER-TEST", "repeats": 1},
```
luca-belli: @jgieringer for viz. not trying to add this yet, but could be interesting

scripts/run_recommended_vera_pipeline.sh
luca-belli: understood

configs/recommended-SI.json
```diff
@@ -0,0 +1,25 @@
+{
```
luca-belli: then make the change!

vera_cli/pipeline.py
luca-belli: but if we make then optional, aren't we doing it wrong for the judge that now could accept an incomplete config?

vera_cli/pipeline.py
luca-belli: add in the TODO a line about figuring out if this is nedded at a later stage

vera_cli/pipeline.py
luca-belli: defer --target all later once we have a concrete usecase. would that break anything?

vera_cli/pipeline.py
luca-belli: then add commebnts

vera_cli/pipeline.py
luca-belli: add comment to explain

vera_cli/pipeline.py
luca-belli: is there a way of reading the defaults from CLI and applying the same logic to config?

vera_cli/pipeline.py
luca-belli: add a doc file for pipeline explaining that many things ca't be guessed at the beginning, such as the folder paths, so it's a different case from the other ones

vera_cli/pipeline.py
luca-belli: that feels wrong. are there any other options?

vera_cli/pipeline.py
luca-belli: if this part is to only enable target all then we should drop it at this time

vera_cli/pipeline.py
luca-belli: and isn't there a parser specific file?

generate.py
luca-belli: ok thanks

scripts/run_recommended_vera_pipeline.sh
luca-belli: Done in 6efe3b3 — the script is back, byte-for-byte except for a note in its header.

I kept it verbatim rather than turning it into a wrapper around `vera pipeline --config configs/recommended-SI.json`. Reasoning, now that I've written both: a wrapper removes the duplication but also removes the rollback path, which is the reason for keeping the script at all. If `vera pipeline` has a bug, a wrapper has the same bug. So the note in the header instead:

> Second source of truth, on purpose and temporarily: the same published profile is also checked in as ../configs/recommended-SI.json […] Nothing keeps the two in sync […] Change one and change the other, until this script is retired in favor of `vera pipeline`.

Say the word if you'd rather have the wrapper and accept that trade.

Everything that followed from the deletion is undone with it: the CHANGELOG "Breaking / migration" entry is gone entirely (with the script staying, `vera pipeline` is purely an addition), README presents the two as two ways to run the same profile, and the `VERA_*` table is back in the present tense as "script variable → `vera pipeline` equivalent".

configs/recommended-SI.json
luca-belli: Done in 9e55ebc — `configs/recommended.json` → `configs/recommended-SI.json`, with the references in `README.md`, `CHANGELOG.md`, `AGENTS.md` and `tests/unit/test_vera_pipeline.py` updated.

configs/recommended-SI.json
luca-belli: Understood — leaving it out of this PR. The placeholder and the `jq` line in the README stay as they are for now; nothing here forecloses the override later, since adding `-c` to the config path only widens what's accepted.

vera_cli/pipeline.py
luca-belli: Good catch, and it's the thing that decides whether the follow-up is worth doing.

It would be wrong if the dataclass were the only gate. It isn't: `vera judge` already validates its own input before constructing anything — `_reject_target_all`, `_output_root` refusing to guess a location for a flat folder, `required()` for `max_concurrent` and `per_judge`. So the fix is to move the requirement to the command that has it rather than dropping it: `utils/config_schema.JudgingConfig` describes what the type can hold, and `vera_cli/judge.py` keeps requiring `conversations` when it resolves input. Judge's error message doesn't change, and it stays impossible to run `vera judge` without conversations.

The split is "what is representable" vs "what this command demands", and it's the one the schema already uses elsewhere — `ScoringConfig.output` is `None`-able because "write beside the input" is a real state, not because scoring is lax about it.

If that separation doesn't hold up in review, the alternative is a separate `PipelinePlan` type in `vera_cli/` holding the not-yet-resolvable fields, leaving `utils/config_schema.py` untouched. More types, same result, no shared dataclass loosened.

I've written this up in the new `docs/pipeline.md` (the "Known follow-up" section) rather than only in this thread, so the decision isn't lost. Still a follow-up, not this PR.

vera_cli/pipeline.py
luca-belli: Done in 9e55ebc:

```python
# Debug is per invocation, not per stage: `set_debug` flips a process
# global (`utils/debug.py`) and `debug` lives on `InvocationConfig`, which
# is one per run.
# TODO: decide later whether per-stage debug is worth having — a
# `--debug-stage generate|judge|score` flag toggling this global around
# each stage would cost ~10 lines and no schema change, but nothing needs
# it yet. Revisit once someone has had to debug one stage of a long
# pipeline; drop this TODO if that never happens.
```

vera_cli/pipeline.py
luca-belli: Done in 9e55ebc, and no — it breaks nothing that exists.

`vera pipeline --target all` now errors on both the flag path and the config path, with `pipeline does not support --target all: run one target at a time.` Nothing shipped used it: the command is new in this PR, only one target (`SI`) is checked in, and `--target all` keeps its full meaning for `vera generate`, which is untouched.

What it deleted is the reason I'm glad you asked: the second rubric-resolution path, which walked back from a run's persona file to `manifest.json` to find which target produced it. That existed only for `all`, and per your comment below it's gone with it. What's left delegates to `judge._rubric_files`.

Widening it later is additive — resolve one pipeline per target instead of rejecting — so deferring costs nothing. Recorded in `docs/pipeline.md` and in the note in `docs/vera-cli-use-cases.md` next to the judge deferral.

vera_cli/pipeline.py
luca-belli: Done in 9e55ebc — both arms, saying why rather than restating the control flow:

```python
# Not dead code, and the order matters: `ConfigError` *is* a `ValueError`
# (`vera_cli/config.py`), so without this arm the broad one below would
# catch every error the resolvers already diagnosed and wrap its message a
# second time — "judging config is missing required field: models" would
# reach the user as "invalid pipeline config: judging config is missing
# required field: models".
except ConfigError:
    raise
# Raw TypeError/ValueError come from the dataclass constructors underneath,
# which have no idea they are behind a CLI — `ModelSpec.from_dict` on a
# wrong-typed field is the common one. Normalized so the user sees a config
# error rather than a traceback.
except (TypeError, ValueError) as error:
    raise ConfigError(f"invalid pipeline config: {error}") from error
```

vera_cli/pipeline.py
luca-belli: Done in 9e55ebc:

```python
# Both checks run before `generate._from_cli`, deliberately: that call
# resolves a whole generation config — target resolution, manifest I/O —
# and would report its own missing flags first, so a `vera pipeline -c X
# -u Y` with no `-j` would do all that work and then complain about
# something else. Checking here also has no "below" that could replace it:
# `generate` has no concept of judge models, so if this does not check
# `-j`, nothing does.
```

vera_cli/pipeline.py
luca-belli: Mechanically yes, and it's two lines — `judge_command.DEFAULTS` is a plain module-level dict, so `_from_config` could do `judging.get("max_concurrent", judge_command.DEFAULTS["max_concurrent"])` instead of `required(...)`.

I'd rather not, and not just for pipeline: it would be a repo-wide change to what a stored config *means*. Today a config is a complete description of a run, so a file written six months ago produces the same run today. Once config mode falls back to `DEFAULTS`, changing `DEFAULTS["max_concurrent"]` from 10 to 20 silently changes every checked-in config that omitted it, including `configs/recommended-SI.json` — the published profile would move without a diff. The same reasoning is why `vera judge` uses `required()` for exactly these two fields (`vera_cli/judge.py`), and pipeline matching it is the consistency, not the duplication.

The cost is real but small and one-time: writing a config by hand means stating two extra fields, and `--print` emits them for you, fully resolved.

If you'd like defaults in config mode anyway, the change that keeps the guarantee is to apply them at *write* time rather than read time — `--print` already materializes every default, so any config produced by the tool is complete. Happy to take that as a follow-up; I don't think it should land as a pipeline-only exception.

vera_cli/pipeline.py
luca-belli: Done in 9e55ebc — new `docs/pipeline.md`, linked from `AGENTS.md`, `vera_cli/README.md`, `CHANGELOG.md`, and the module docstring.

It makes exactly the point you're asking for: a single-stage command is handed everything it needs, but a pipeline *produces* two of its own inputs —

```
generate ──▶ <run>/conversations ──▶ judge ──▶ <evaluation>/results.csv ──▶ score
```

— so `judging.conversations` and `scoring.results` can't be stated before the run starts; the folder doesn't exist and its name is minted by generation. Which is why a pipeline config states **fewer** fields than a judge config, and why stating `judging.conversations` is an error rather than an override.

It also covers what that costs in the code today (the loose fields on `PipelineRun`, the hand-built `to_dict`), the follow-up that would remove the cost, and the `--target all` deferral.

vera_cli/pipeline.py
luca-belli: Agreed it feels wrong — it's the same smell as your `to_dict` comment, and it's worth listing what's actually available rather than just deferring it again.

1. **Make `JudgingConfig.conversations` / `ScoringConfig.results` optional, keep the requirement in `vera_cli/judge.py`.** `PipelineRun` collapses to three real config objects; the unpacking here, the hand-built `to_dict`, and the reconstruction in `_judge_and_score` all go. Your concern about judge accepting an incomplete config is answered in that thread — the requirement moves to the command, not away. **My recommendation.**
2. **A `PipelinePlan` type in `vera_cli/`** holding the not-yet-resolvable judging/scoring fields as a named thing, with a `with_conversations(folder) -> JudgingConfig` method. Same shape as today but the looseness has a name and one place to complete it, and `utils/config_schema.py` is untouched. Cheapest change; least payoff.
3. **Build the `JudgingConfig` with a placeholder `conversations`** and overwrite it per run folder. Rejected — a config object that is briefly a lie is worse than fields that are honestly absent, and it would round-trip the placeholder through `to_dict`.
4. **Give `JudgingConfig` a sibling `JudgingSpec` without `conversations`,** with `JudgingConfig = JudgingSpec + conversations`. Most correct typing, most machinery, and every consumer has to learn which one it's holding.

(1) and (2) are both small; (1) is the one that deletes code rather than relocating it. Either way it's a follow-up — it touches `utils/config_schema.py` and the commands validating those fields — and `docs/pipeline.md` now records the reasoning so the next reader doesn't have to reconstruct it from this thread.

vera_cli/pipeline.py
luca-belli: Dropped in 9e55ebc. That was exactly what it was for.

`_rubric_for` is now three lines and no branch:

```python
def _rubric_for(target: str) -> RubricFiles:
    return judge_command._rubric_files(load_target(resolve_target_manifest(target)))
```

The `assert generation is not None` and the walk back from `personas[0]` to `manifest.json` are gone with the `all` support, and the hand-rolled `RubricFiles(...)` tail is now judge's projection.

vera_cli/pipeline.py
luca-belli: There isn't one — I checked again rather than trusting the earlier answer.

`vera.py` is the whole of the parser assembly: it builds the root parser, creates the subparsers, and calls each command module's `register`. The `vera_cli/` package is `config.py`, `targets.py`, and one module per command (`generate`, `judge`, `score`, `pipeline`); no `parsers.py`, `cli.py`, or `args.py`. Each `register()` lives at the bottom of its own command's module, and pipeline's is the fourth instance of that pattern.

If you'd like the flags to live somewhere separate, that's a repo-wide move worth doing as its own change — `vera generate` is the reference implementation the other three follow, so pipeline should either match it or the convention should change for all four.

docs/pipeline.md
```diff
@@ -0,0 +1,88 @@
...
+}
+```
+
+`--print` round-trips this form: it emits a valid `vera pipeline --config`
+document, which is deliberately *not* a valid `vera judge` one.
+
+## What that costs in the code today
+
+`vera_cli/pipeline.py` resolves in two halves. Everything knowable before a
+model is called still is — a bad rubric or an unknown target fails before any
```
luca-belli: use chicago style for em dashes, so without spaces. make sure to add the proper dash as well

vera_cli/pipeline.py
luca-belli: otherwise could the dataclass have another class that has `is_pipeline` and if not true needs to specify all the values?

vera_cli/pipeline.py
luca-belli: 1 seems promising, but it feels wrong to have a config not declaring something needed, just to redeclare it needed late

is there anything we could do using inheritance? eg JudgingConfig inshrits from JudgingSpec or viceversa

vera_cli/pipeline.py
luca-belli: add a todo item to move the parser in their own files

docs/pipeline.md
luca-belli: Done in 68342f0. Every em dash in `docs/pipeline.md` is now closed up (`property—the stages are chained—makes`) and uses U+2014. The one that fell at a line break is rewrapped, because a trailing `chained —` would render as `chained— makes`.

I kept this to the doc. Code comments across `vera_cli/` still use spaced dashes, and changing those is a repo-wide pass rather than something for this PR.

vera_cli/pipeline.py
luca-belli: I'd avoid the flag, and the inheritance in the other thread gets you what you want here without it. Done that way in 68342f0.

The trouble with `is_pipeline` is that the type stops telling you whether `conversations` is there. Every place that reads a `JudgingConfig` has to check the flag first, and forgetting to check fails at runtime, far from where the config was built. It's option 1 (optional fields) with an extra branch.

With the split, `JudgingConfig` still requires every field, `vera judge` builds one exactly as before, and there's no way to hand it an incomplete one. The pipeline holds a `JudgingSpec` instead, which is a different type.

vera_cli/pipeline.py
luca-belli: Yes, and it's better than all four options I listed. Done in 68342f0, `JudgingConfig` inheriting from `JudgingSpec` (not the other way round):

```
JudgingSpec     models, rubrics, max_concurrent, per_judge
  └─ JudgingConfig  + conversations, output

ScoringSpec     personas, skip_risk_analysis
  └─ ScoringConfig  + results, output
```

- **Nothing becomes optional.** `JudgingConfig` still requires `conversations` and `output`, which answers your objection to option 1. `vera judge` / `vera score` are untouched.
- **Why this direction.** A config is a spec plus two fields, so it can stand in wherever a spec is expected. The reverse would mean a subclass with *fewer* required fields than its parent, which breaks substitution.
- **`complete(conversations=..., output=...)`** on the spec returns the full config. That's the only step `_judge_and_score` needs once the run folder exists.

What this removes from `vera_cli/pipeline.py`:

```python
@dataclasses.dataclass(frozen=True)
class PipelineRun:
    generation: RunConfig
    judging: JudgingSpec
    scoring: ScoringSpec
```

The unpacking is gone, `to_dict` delegates to all three, and the `JudgingConfig(...)` rebuild in `_judge_and_score` is now `pipeline_run.judging.complete(...)`.

It also fixed a bug. Before, `judging.max_concurrent` and `judging.per_judge` in a pipeline config were only type-checked when the `JudgingConfig` was built, **after generation had already run**. So `"max_concurrent": "ten"` spent the whole generation budget and then failed. The spec validates at resolve time now. There's a new test for that, and I confirmed it fails on the previous commit. `docs/pipeline.md` now describes this design and why the optional-field and flag options were rejected.

vera_cli/pipeline.py
luca-belli: Added to `TODO` in c6b0d013. It covers all four commands in one change, since `vera generate` is the one the others copy:

> Move each `vera` subcommand's argparse `register()` out of its command module (`vera_cli/generate.py`, `judge.py`, `score.py`, `pipeline.py`) into a parser-only module. Do all four in one change, since `vera generate` is the reference the other three copy.

vera_cli/pipeline.py
luca-belli: Follow-up: I first read "the TODO" as an inline comment. It should have been the repo's `TODO` file, so it's there now (c6b0d013), and the inline block in `run()` is down to one line pointing at it.

vera_cli/pipeline.py
luca-belli: document the difference between judging spec and  config in an appropriate file. don't be too long

utils/config_schema.py
```diff
@@ -303,24 +303,25 @@ def to_dict(self) -> dict[str, Any]:
 
 
 @dataclasses.dataclass(frozen=True)
-class JudgingConfig:
-    """What judging run to perform — the run-defining half for `vera judge`.
+class JudgingSpec:
```
luca-belli: add (or add to an existing) a doc (MD file) to explain this difference

luca-belli: Done in 6efafdb: [`docs/design/judging-and-scoring-specs.md`](https://github.com/SpringCare/VERA-MH/blob/feat/vera-pipeline/docs/design/judging-and-scoring-specs.md).

I put it under `docs/design/` rather than extending an existing doc because `utils/config_schema.py` is a stable interface, and `docs/design/README.md` asks for a short design record whenever one changes. It uses that directory's standard format (context, decision, consequences) and is about 60 lines. The core of it is one table:

| Type | Holds | Complete? |
|---|---|---|
| `JudgingSpec` | `models`, `rubrics`, `max_concurrent`, `per_judge` | no: says *how* to judge |
| `JudgingConfig(JudgingSpec)` | spec + `conversations`, `output` | yes: says *what* to judge |
| `ScoringSpec` | `personas`, `skip_risk_analysis` | no: says *how* to score |
| `ScoringConfig(ScoringSpec)` | spec + `results`, `output` | yes: says *what* to score |

After the table it covers who holds which type (only `vera pipeline` holds specs), why the config subclasses the spec, and the two rejected options (optional fields, `is_pipeline`). The `JudgingSpec` docstring now links to it. `docs/pipeline.md` links to it too, and I removed the alternatives paragraph there so the rationale lives in one place.

vera_cli/pipeline.py
luca-belli: Covered by the same record, 6efafdb: [`docs/design/judging-and-scoring-specs.md`](https://github.com/SpringCare/VERA-MH/blob/feat/vera-pipeline/docs/design/judging-and-scoring-specs.md). Details are in the thread on `utils/config_schema.py`.

utils/config_schema.py
```diff
@@ -398,30 +441,64 @@ class ScoringConfig:
...
+    results: str
+    output: str | None
+
+    def __post_init__(self) -> None:
+        if not isinstance(self.results, str) or not self.results:
+            raise ValueError("scoring.results must be a path")
+        if self.output is not None and (
+            not isinstance(self.output, str) or not self.output
+        ):
+            raise ValueError("scoring.output must be null or a path")
```
jgieringer: this seems confusing as we're saying needs to be a path but checking for str

vera_cli/pipeline.py
```diff
@@ -0,0 +1,514 @@
...
+    )
+    transcripts_dir, _, folder_name = resolve_conversation_input(
+        judging.conversations[0]
+    )
+    _, evaluation_folder = await run_judging(
+        judge_models={model.name: model.repeats for model in judging.models},
+        rubric_file=judging.rubrics[0].rubric_file,
+        rubric_prompt_beginning_file=judging.rubrics[0].rubric_prompt_beginning_file,
+        question_prompt_file=judging.rubrics[0].question_prompt_file,
+        transcripts_dir=transcripts_dir,
```
jgieringer: nit: accessing `[0]` feels scary and like I could mess something up, but I guess there are checks across the code the enforce 1 item?

jgieringer (approved):
LGTM! Some nits.
I also wonder if integration tests should be added that run like 2 turns over 1 persona across multiple models per users & judges with the cheapest models like claude haiku 4.5 & gpt 6 luna (or gpt-5.4-nano)

luca-belli:
> LGTM! Some nits. I also wonder if integration tests should be added that run like 2 turns over 1 persona across multiple models per users & judges with the cheapest models like claude haiku 4.5 & gpt 6 luna (or gpt-5.4-nano)

Great idea, will add!

vera_cli/pipeline.py
luca-belli: Good point, I think there was another similar problem on the judge one, but i'll add a comment

utils/config_schema.py
luca-belli: Fair: the check is only for a non-empty string, so "must be a path" promised more than it checked. In b845398 the five messages in `utils/config_schema.py` now say "must be a non-empty path string" (or "null or a non-empty path string").

Whether the file exists is checked earlier, by the command that resolves the path (`vera_cli.config.config_path` → `existing_file`, and `vera_cli/score.py` for `--results`), so a missing file fails with the config field or flag that named it. The schema only sees the resolved string. The module docstring now says this in one sentence, which is the same split `RubricFiles` already documents.

vera_cli/pipeline.py
luca-belli: Followed up in b845398. The checks do exist (`JudgingConfig.__post_init__` requires exactly one conversations folder and one rubric), but you had to go find them. Both call sites, here and in `vera judge`, now unpack instead of indexing:

```python
# Exactly one of each, enforced by `JudgingConfig.__post_init__`; the
# unpacking fails loudly rather than judging only the first if that changes.
(conversations,) = judging.conversations
(rubric,) = judging.rubrics
```

If the length-1 rule is ever lifted, this raises instead of quietly judging only the first folder. `judging.models[0].extra_params` stays, because a run can have several judges. It has a comment now: every judge must share one parameter dict (`JudgingSpec.__post_init__` rejects anything else), so the first stands for all.

luca-belli:
@jgieringer the integration test is in 47cabf1: `tests/integration/test_vera_pipeline_live.py`. It runs `vera pipeline` end to end with 1 persona and 2 turns:

- **users:** `gpt-5.4-nano` and `claude-sonnet-4-5`, so two generation runs and two evaluations
- **judges:** `gpt-5.4-nano` and `gpt-5.6-luna`, each checked to appear in every `results.csv`
- **chatbot:** `gpt-5.4-nano`

It also checks the scores and the "folders to pool" report. It is marked `live`, so CI and `-m "not live"` skip it. It passed against the gateway in about 45s:

```bash
uv run pytest -m live tests/integration/test_vera_pipeline_live.py
```

Two deviations from your list, both forced by the gateway catalog:

- **No Haiku.** The gateway does not serve any Haiku model, so the Claude user model is `claude-sonnet-4-5`, the cheapest Claude it lists.
- **Claude is not a judge.** Through the gateway, Claude answers the rubric prompt in plain text (`ANSWER: No / REASONING: …`) instead of the JSON schema, so every question fails. A trivial prompt does return valid JSON on the same models, so this is specific to the judge prompt and independent of `vera pipeline`. I left it out of this PR; it deserves its own issue.

The model IDs are pinned in the test. Once #223 lands they move to `Config`, so its availability check covers them (tracked in `TODO`).

PR #215, opened 2026-09-15
Docs: target at the root of the output layout, and a phase for vera pool

luca-belli:
Docs only — nothing implements the `c_`/`u_`/`j_` layout yet, so there is no code or output change here. Two commits: the layout itself, then the migration-table corrections it implies.

**Needs a real decision, not a docs rubber-stamp:** this revises adopted rules in `ARCHITECTURE-SPINE.md` — the flat-layout removal rule, the run-root collision check, and the naming row.

## 1. `output/<target>/` becomes the outermost segment

```
output/
  <target>/                                       # scores never cross this boundary
    c_<chatbot>/
      u_<user-model>_<sha>_<timestamp>/
        conversations/
        evaluations/
          j_<judge>_<sha>_<timestamp>/
      pooled/
        u_<a>+<b>_<sha>_<timestamp>/
    evaluations/<config-sha256>/                  # standalone judging
  pooled/<config-sha256>/                         # pooling across chatbots
```

Scores from different targets are never comparable, so the layout makes mixing them structurally impossible rather than merely discouraged, and `ls output/SI/` answers "which chatbots have been evaluated against SI".

This does **not** contradict "generation has no knowledge of rubrics". A target is personas, prompts, *and* a rubric, and generation already consumes the persona files and context template — so conversations genuinely belong to the target that produced them. What stays true is that a run's conversations live together regardless of which rubrics later judge them.

**The `evaluations/<rubric_name>/` segment is therefore gone** — the path already names the target.

## 2. The generated nickname is retired

It existed so a person could refer to a run without quoting a sha (`prophetic-bullfrog_20260713-1530_a1b2c3`). The model name does that better and is also what a reader wants to know, so the default label *is* the model name.

Run-id is now `<model>_<sha>_<timestamp>`, where `<model>` is whichever model the surrounding path does not already name — the user model for a generation run (`c_<chatbot>/` gives the chatbot), the judge for a judging run, the `+`-joined set for a pooled one. An explicit human label may still be supplied, replacing that segment rather than adding to it.

This closes a real gap: `-u gpt-5.2:1 claude-opus-4-5:1` previously produced two sibling run folders distinguishable only by timestamp, because nothing in the path named the user model — the same attribution problem used to justify deferring `vera judge --target all`.

**The timestamp moves last, uniformly** rather than per artifact kind. Everything sharing a prefix still sorts chronologically, and one ordering rule beats one per artifact.

## 3. Pooled output gets a home it didn't have

Previously a pooled result had nowhere to live in this scheme: it spans two runs, so it belongs under neither, and `scripts/pool_vera_scores.py` invents a synthetic legacy-style `j_<judge>__p_<a>+<b>__a_<chatbot>…` folder — the exact naming Phase 3 retires.

Now it is `c_<chatbot>/pooled/`, a persistent container in the same role as `c_<chatbot>/` itself — not a run-root, not collision-checked. The user-model combination is named in the path deliberately, so listing one directory says which pools exist without opening a metadata file; bounded at three names before `+Nmore`, with `pool_metadata.json` authoritative for exact sources. Sources spanning more than one chatbot group under a top-level `pooled/<config-sha256>/`, the same nested-versus-standalone split judging already makes.

Also now documented, having been true only by accident: **pooling writes only the merged `results.csv`, `pool_metadata.json`, and `scores/`** — transcripts and per-question TSVs are never copied.

## 4. Migration-table corrections (second commit)

**`vera pool` had no phase.** It is described as first-class at `architecture.md:48`, `:147`, and `:357`, but no row shipped it: Phase 1's Done-when lists every other command flag and omits `pool`; Phase 2 is explicitly "pure move, no new behavior" so it cannot introduce a subcommand; and Phase 5 said "if `vera pool` has already fully superseded it by now, just delete it" — resting on a command no earlier phase promised.

Phase 1 now ships it. That phase's goal is that `vera.py` replaces the top-level scripts, and `scripts/pool_vera_scores.py` is one of them — leaving pooling script-only made "`vera.py` is the only documented entry point" false for the one capability the recommended profile depends on. Phase 2 relocates the implementation into `score/pool.py` without touching the command surface, and Phase 5's check stops being conditional: the script is dead code by then and is simply deleted.

**Where the layout change lands.** Phase 3 was already rewriting `utils/naming.py`, so it carries the `<target>/` root and `c_<chatbot>/pooled/`. Because the target root attributes judging output by itself, Phase 4 never builds the per-rubric segment and no longer enables `vera judge --target all` — Phase 3 does, since attribution was its only blocker. Both Done-when bars are updated; Phase 4's previously required "separated `evaluations/<target>/` output", which this design does not produce and would have left the bar unsatisfiable.

Net effect on sequencing: Phase 4 shrinks to `rubrics[]` length > 1 plus per-rubric judge-model overrides. Dependency order is unchanged — 3 still precedes 4, and 4 still needs 3's naming.

## Known risk: mixed component selection

`--personas A --rubric B` generates from A and judges with B, so the run lives under `output/A/` with nothing in the path recording B's rubric. This stays **supported and intentionally unhandled at the naming level** — it is an expert path and does not warrant a conditional path segment.

What the docs now say, rather than filing it as a gap, is what is actually at stake: the target root exists so that scores never cross it, and a mixed run breaks that guarantee *inside* a root. Anything locating results by globbing a target root — `vera pool` above all — could merge two rubrics' scores without noticing.

The mitigation is verification, not naming. Aggregation must establish rubric identity from each source's own recorded config and refuse to merge sources that disagree, never infer it from the enclosing path. Two existing `TODO` entries already call for exactly that capability: persisting rubric dimensions plus a rubric fingerprint with evaluation output, and recording the shas of the files a config consumed. **Until those land, a mixed run is safe to produce and unsafe to pool blindly** — which is a constraint `vera pool` will need to honour.

## Relationship to #214

#214 drops its implicit pooling stage on the strength of this PR: `vera pipeline` stays the three stages `architecture.md:148` names, and pooling becomes the explicit `vera pool` that Phase 1 now promises.


docs/vera-cli-use-cases.md
```diff
@@ -333,28 +338,58 @@ Each list entry's `name` is always a **specific model identifier** (e.g. `claude
...
+
+- **Generation** groups persistently by chatbot model (`c_sonnet/` accumulates every run against that model). **Judging** stays flat per-run — intentionally asymmetric, not an inconsistency.
+
+- **Every run-id is `<model>_<sha>_<timestamp>`**, where `<model>` is whichever model the surrounding path does not already name: the *user* model for a generation run (the chatbot is given by `c_sonnet/`), the *judge* model for a judging run, and the `+`-joined set of user models for a pooled result. Readable (which model), integrity-checked (sha256 of `config.json`), ordered (when).
+
+  **The timestamp goes last, uniformly.** Everything sharing a prefix still sorts chronologically, and one ordering rule is easier to hold than one-per-artifact-kind.
+
+  **There is no generated nickname.** An earlier draft used a human-memorable tag (`prophetic-bullfrog`) so a person could refer to a run without quoting a sha. The model name does that job better, because it is also the thing a reader wants to know, so the default *is* the model name and the arbitrary tag is retired. An explicit human label may still be supplied per run; it replaces the model segment rather than being added to it.
```
luca-belli: add the option to add a human name. defaults to none and has model name

emily-vanark (approved):

docs/vera-cli-use-cases.md
luca-belli: Done in `ad05942d` — you're right that additive is the better call, and my wording had it backwards.

The label is now *added* rather than substituted:

```text
u_<user-model>[_<label>]_<sha>_<timestamp>
```

It **defaults to none**, so the plain form stays the model name alone. Replacing the model was a bad trade — a run labelled `smoke-test` would no longer say what it ran against, which is the one thing a reader always wants. The model also stays first so listing a directory still groups by model.

Two things the earlier text left open, now pinned down:

- **The label is run-defining**, so it lives in the run's `config.json` and participates in the config sha like every other field. Two otherwise-identical runs with different labels are therefore different runs — which is what labelling them separately asserts.
- **It is deliberately not spelled `--run-id`** (legacy `generate.py`'s flag). It is a decorative handle, never an identifier; the sha remains the identity.

The only piece I did *not* decide is the flag name — the doc describes the behaviour and calls it a label, but doesn't fix the spelling. `--label` is my inclination over `--name` or `--tag`, since "name" invites confusion with the model `name` field that already exists in `ModelSpec`. Happy to pin it either way.

nz-1 (approved):

docs/vera-cli-use-cases.md
luca-belli: Flag spelling deferred, and recorded as such in `4a227948` rather than left implicit — since this doc is marked `status: resolved`, an unnoted open question inside it would read as a decision already taken.

The bullet now says the semantics are settled (optional, additive, defaults to none, run-defining, sha-participating) and that only the CLI flag and config-field names are open, until Phase 3 implements the label. That phase rewrites `utils/naming.py` (needs the flag) and formalizes the config shape in `utils/config_schema.py` (needs the field), so naming either earlier would just mean naming it twice.

PR #218, opened 2026-09-16
fix: surface real CLI failure causes in live pipeline tests

luca-belli:
## Summary

The live `TestVERAMHPipeline` tests were failing with a message that said nothing at all:

```
RuntimeError: Generate CLI failed:
```

The cause was laundered through three layers, each of which dropped it:

1. `generate_conversations/runner.py` catches a per-conversation failure, records `skipped=True`, and prints the real cause to **stdout**.
2. `generate.py`'s all-skipped branch then exits 1 with **nothing on stderr**.
3. The test helpers reported only `result.stderr` — hence the bare label. The in-process `run_pipeline` test surfaced a raw `SystemExit: 1` instead.

So any failure of any kind — bad key, unreachable endpoint, retired model — produced the same uninformative string. This PR makes the failure legible and fixes two correctness issues found on the way.

**Observability**
- `cli_failure()` helper reports the exit code plus stdout, replacing the two hand-rolled stderr-only messages.
- `generate.py`'s all-skipped exit prints the distinct skip reasons and their errors to stderr, so "every conversation already existed" and "every conversation errored" are no longer indistinguishable.

**Correctness**
- Two `runner.py` error paths recorded `skipped=True` with an `error` but no `skip_reason`, so the end-of-run `"(N skipped due to errors)"` tally undercounted them and they reported as `skip_reason=unknown`. Both now set `skip_reason="error"`.
- `validate_test_environment` reads `os.getenv` for API keys, but those are loaded by `load_dotenv()` at import of `llm_clients/config.py` — so whether the guard could see them depended on pytest's collection order. It now calls `load_dotenv()` itself, and a keyless machine skips instead of failing deep inside a subprocess.
- `test_pipeline_error_handling` asserted `match="Generate CLI failed"` — a prefix every failure carries, so unrelated breakage passed it. It now matches the unsupported-model error it actually feeds.

**Retired model**
- `TEST_CONFIG["PROVIDER_MODEL"]` pinned `claude-opus-4-1-20250805`; Claude Opus 4.1 retired 2026-08-05. Replaced with `claude-opus-5`. This was a latent failure rather than the one reported above.

## Architecture checklist

- [ ] Read [architecture.md](docs/architecture.md)
- [ ] OpenSpec change linked (required for multi-file or cross-boundary work) — n/a, no boundary changes
- [x] `uv run pyright` — no new errors (the 4 pre-existing `reportArgumentType` errors in `test_scoring.py` are unchanged from `main`)
- [x] `uv run pytest -m "not live"` passes — 1101 passed, 8 deselected
- [ ] Label **`architecture-change`** — not needed

## Test plan

- [x] Local checks run: `pytest -m "not live"`, `ruff check`, `ruff format --check`, `pyright`
- [ ] CI green
- [ ] Live suite (`pytest tests/integration/test_scoring.py -m live`) — **not verified here.** It needs reachable Anthropic and OpenAI endpoints; on my machine Anthropic calls fail at the transport layer for environmental reasons unrelated to this diff. Worth a run by someone whose environment reaches both. Note that the observability changes are what made that distinction visible in the first place.


tests/integration/test_scoring.py
```diff
@@ -26,9 +29,29 @@
 }
 
 
+def cli_failure(label: str, result: subprocess.CompletedProcess) -> RuntimeError:
```
luca-belli: if this is extending an error, should it not inherit from Runtime Error?

tests/integration/test_scoring.py
```diff
@@ -9,13 +9,16 @@
...
 import pytest
+from dotenv import load_dotenv
 
 # Import additional modules to increase test coverage
 
 # Test configuration constants
 TEST_CONFIG = {
     "USER_MODEL": "gpt-4o",
-    "PROVIDER_MODEL": "claude-opus-4-1-20250805",
+    # claude-opus-4-1-20250805 retired 2026-08-05; claude-opus-5 is the
```
luca-belli, line 19: worth saving all those models config somewhere so they percolate automatically?

tests/integration/test_scoring.py
luca-belli: Agreed, and fixed in 49baf805 — it's now a class rather than a factory function.

To be precise about what it was: `cli_failure` wasn't extending anything, it was a function *returning* a `RuntimeError` (hence `raise cli_failure(...)` at the call sites). But your instinct points at the better design, so:

```python
class CliFailure(RuntimeError):
    def __init__(self, label: str, result: subprocess.CompletedProcess) -> None:
        self.label = label
        self.result = result
        super().__init__(...)
```

Two things this buys beyond reading better:

- Callers can catch `CliFailure` specifically instead of `RuntimeError`, which in this file also covers "couldn't parse the folder path" and "generated directory is not valid" — different problems that currently can't be told apart by type.
- Keeping `result` on the instance means a test can assert `exc.result.stdout` rather than regex the rendered message. That matters for `test_pipeline_error_handling`, which today matches a string.

Subclassing `RuntimeError` keeps every existing `pytest.raises(RuntimeError)` working, so nothing else had to change.

tests/integration/test_scoring.py
luca-belli: Yes — and the interesting part is that the central home already exists and is dead.

`llm_clients/config.py` has `Config.get_claude_config()` / `get_openai_config()` / `get_gemini_config()`, which look exactly like the thing you're describing. Nothing calls them except their own unit tests:

```
$ grep -rn "get_claude_config\|get_openai_config\|get_gemini_config" --include='*.py' .
tests/unit/llm_clients/test_config.py:22,33,44
llm_clients/config.py:72   (the definition)
```

And `get_claude_config()` currently returns `claude-sonnet-4-5-20250929`, which is itself stale — so the one place that *is* supposed to be the source of truth has been quietly rotting because nothing depends on it.

What the spread actually looks like, having inventoried it:

- **Production code: no functional hardcoded model IDs.** Every hit in `generate.py`, `judge.py`, `run_pipeline.py`, `llm_factory.py` is argparse help text, a docstring, or an `%(prog)s` usage example. The CLIs require the model to be passed explicitly. That's why a retirement doesn't break normal runs — it only breaks callers that pin one.
- **Tests are where the pins live**, and several are well past their expiry: `test_scoring.py` had `claude-opus-4-1-20250805` (this PR), `test_conversation_runner.py` uses `claude-3-opus-20240229`, `test_pipeline.py` and `test_judge_cli_extra_params.py` pin `claude-sonnet-4-5*`. Most are mocked so they never notice; the live ones are the ones that bite.
- **Docs pin too** — `README.md:116` names `claude-opus-4-5-20251101` for the headline pipeline, and `AGENTS.md:114-115` uses `claude-sonnet-4-5-20250929`.

So I'd say the fix isn't "add a config", it's "make the existing one load-bearing, then point the tests and docs at it". Worth adding a `live` smoke test that resolves each configured ID against the provider's models endpoint, so a retirement fails with "this model is gone" rather than as a generic connection error six layers down — which is precisely the failure mode this PR exists to fix.

I'd rather not fold that into this hotfix: it touches four test files, the docs, and a module with its own unit tests. Happy to open it as a follow-up issue if you want it tracked.

luca-belli: ok add a follow up please

luca-belli: Opened as #222.

One correction to what I wrote above, since it changes the shape of the follow-up: I said the `get_*_config()` getters were dead, called only by their own unit tests. That's wrong. The three client constructors call them as fallback defaults —

```
llm_clients/claude_llm.py:173   self.model_name = model_name or Config.get_claude_config()["model"]
llm_clients/openai_llm.py:108   self.model_name = model_name or Config.get_openai_config()["model"]
llm_clients/gemini_llm.py:154   self.model_name = model_name or Config.get_gemini_config()["model"]
```

— and that was already true when I wrote the comment; my grep excluded `llm_clients/*_llm.py` and I read the absence as absence. `LLMFactory` types `model_name: str` and always passes it, so the `or` branch is only reachable via direct construction, which is why the stale `claude-sonnet-4-5-20250929` default has gone unnoticed. Near-unexercised, not unreachable — so it's a narrow live correctness issue, not just rot in unused code. #222 notes this, and adds the question of whether that fallback should exist at all given the factory makes it redundant.

The recommendation is unchanged: make the existing config load-bearing rather than add a parallel one, point the *live* pins at it (mocked tests can keep literals — the ID is arbitrary there), and add a `live` smoke test resolving each configured ID against the provider's models endpoint, so the next retirement fails with "this model is gone" instead of the transport error that #218 exists to make legible.

nz-1 (approved):

